PR #103, opened 2021-09-27
Add token field to reward claim

tintinthong:
reward claim event is missing a token field. this helps us understand what token is being claimed. It is unimportant now because card.cpxd is the only reward token but it will be important later when we add facility for other tokens.

subgraph `RewardeeClaim` doesn't track this. Following a deploy of cardpay protocol to sokol we need to:
- update contract in sdk
- update abi of reward pool in sdk
- add token field into `RewardeeClaim` entity

**BREAKING CHANGE due to change of event structure**

### DIFF
--- a/contracts/RewardPool.sol
+++ b/contracts/RewardPool.sol
@@ -21,6 +21,7 @@ contract RewardPool is Initializable, Versionable, Ownable {
     address rewardProgramID,
     address rewardee,
     address rewardSafe,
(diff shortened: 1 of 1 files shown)


tintinthong:
I guess to make this deploy more seamless it should go in with eip1271 pr

habdelra (approved):
looks good

PR #107, opened 2021-10-12
Upgradeable contracts with deploy and tests

alexspeller:
Make OZ contracts upgradeable, add to deploy script, and add a build step that tests deploy / upgrade


habdelra (approved):
aside from the concern around setting an owner in the commerce protocol contracts, this looks good

PR #117, opened 2021-10-24
Fix import casing error

tintinthong:
When I run `yarn build:clean` I get an error

```
Error HH409: Trying to import CardPay-Contracts/contracts/BaseERC20.sol from contracts/commerce-protocol/Contracts.sol, but it has an incorrect casing.
```

This change fixes the build


tintinthong:
Ooops hmmm thought it worked on my local. Let me check

PR #123, opened 2021-11-01
Make prepaid card test less-stateful

tintinthong:
Wanted to make each test isolated using the `only()` primitive. So now, the test would work with `test.only` or `describe.only`. Previously, some test were dependent on previous test in previous describe blocks running.


habdelra (approved):
Looks great, thx!

PR #129, opened 2021-11-02
Remove program function to use governance admin

tintinthong:
- adds `governanceAdmin` to reward manager
- modifies `removeRewardProgram` to only be called by `governanceAdmin`. So far `removeRewardProgram` is the only governance-level function exposed by the reward manager.


contracts/RewardManager.sol
```diff
@@ -54,6 +54,7 @@ contract RewardManager is Ownable, Versionable, Safe {
   uint256 public rewardeeRegistrationFeeInSPEND;
   uint256 public rewardProgramRegistrationFeeInSPEND;
   address payable public rewardFeeReceiver; // will receive receive all fees
+  address public governanceAdmin; // eoa with governance powers
```
alex-cardstack, line 57: Is the reward manager contract deployed yet? If so, I think adding this here will break storage layout when the implementation is upgraded.

contracts/RewardManager.sol
```diff
@@ -103,6 +109,7 @@ contract RewardManager is Ownable, Versionable, Safe {
     rewardeeRegistrationFeeInSPEND = _rewardeeRegistrationFeeInSPEND;
     rewardProgramRegistrationFeeInSPEND = _rewardProgramRegistrationFeeInSPEND;
     versionManager = _versionManager;
+    governanceAdmin = _governanceAdmin;
```
alex-cardstack, line 112: Do we need a setter so we can update this?

contracts/RewardManager.sol
```diff
@@ -127,7 +134,7 @@ contract RewardManager is Ownable, Versionable, Safe {
     emit RewardProgramCreated(rewardProgramID, admin);
   }
 
-  function removeRewardProgram(address rewardProgramID) external onlyOwner {
+  function removeRewardProgram(address rewardProgramID) external onlyGovernanceAdmin {
```
alex-cardstack, line 137: This now means the owner cannot call the method, _only_ the governance admin.

Does it make sense to have a `onlyOwnerOrGovernanceAdmin` modifier for this purpose?

contracts/RewardManager.sol
habdelra: It will not be upgrade safe, but we are still in the mode of making destructive changes to the rewards contracts since we have no real clients yet.

contracts/RewardManager.sol
habdelra: We’ve been saving on contract size by consolidating the setters into this general setup() function

contracts/RewardManager.sol
```diff
@@ -86,7 +91,8 @@ contract RewardManager is Ownable, Versionable, Safe {
     uint256 _rewardeeRegistrationFeeInSPEND,
     uint256 _rewardProgramRegistrationFeeInSPEND,
     address[] calldata _eip1271Contracts,
-    address _versionManager
+    address _versionManager,
+    address _governanceAdmin
```
habdelra, line 95: Not a super big deal, but I’ve been trying to keep the version manager address as the last param in the setup function for consistency sake

habdelra (approved):
Looks good

contracts/RewardManager.sol
habdelra: When @tintinthong is ready for the SDK updates, the subgraph will be updated to start indexing this contract  at the block the new reward manager is deployed

contracts/RewardManager.sol
tintinthong: I wud think to keep it simple. The owner is the deployer. The governance admin is a dao or an eoa under our control.  There is nothing stopping the the owner to configure the governance admin to point to itself. This edit wud just be purely for convenience -- don't think we need that.

alex-cardstack (approved):

contracts/RewardManager.sol
tintinthong: there are quite a few changes happening at the same time. I will bottle up all those in a single sdk update.

PR #130, opened 2021-11-02
Add blob as rule

tintinthong:
- pr is mainly just adding blob as rule to card protocol
- consideration of size of bytes passed into function. Turns out the based on empirical analysis the bytes parameter does not overkill gas as discussed [here](https://linear.app/cardstack/issue/CS-2350/size-of-bytes-for-proof-and-addrewardrule)
- added unhappy test for `removeRewardProgram`. All functions cannot be executed after `removeRewardProgram` is called except `transferRewardSafe`
- change `rewardProgramLock` to a toggle rather than a switch off-only function
- remove `removeRewardRule`. We just want the `governanceAdmin` to interact by removing reward program -- easy and simple.


contracts/RewardManager.sol
```diff
@@ -145,26 +135,18 @@ contract RewardManager is Ownable, Versionable, Safe {
...
-  }
-
-  function removeRewardRule(address rewardProgramID, string calldata ruleDID)
+  function addRewardRule(address rewardProgramID, bytes calldata blob)
     external
     onlyHandlers
   {
-    delete rule[rewardProgramID][ruleDID];
-    emit RewardRuleRemoved(rewardProgramID, ruleDID);
+    rule[rewardProgramID] = keccak256(abi.encodePacked(blob));
```
habdelra, line 142: i thought the intent was to store the rule blob on chain, not a hash of it, no?

tintinthong: thats actually a good point. @IanCal do you have an opinon on this?

tintinthong:
@habdelra is this good to merge?

habdelra (approved):
👍

PR #131, opened 2021-11-03
Initial withdrawal implementation

alex-cardstack:
Based on swap owner implementation

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,6 +115,10 @@ The `RewardManager` contract is responsible for creating the gnosis safes that a
 1. The `rewardee` or the owner of the prepaid card that executed the `registerRewardee` action.
 2. The `RewardManager` contract itself.
 
+Withdraw is only possible to the EOA owner of the reward safe, using the `withdrawFromRewardSafe` action with a wrapped gnosis signature, the inner signature being for the `token.transfer(to, value)` call, and the outer signature being for the `withdrawFromRewardSafe` function.
+
+Only approved tokens can be withdrawn from the reward safe.
+
 ## Prerequisites
 The following prerequisites are required for this project:
 - NodeJS ver 14+
--- a/contracts/RewardManager.sol
+++ b/contracts/RewardManager.sol
@@ -26,6 +26,7 @@ contract RewardManager is Ownable, Versionable, Safe {
     address oldOwner,
     address newOwner
   );
+  event RewardSafeWithdrawal(address rewardSafe, address token, uint256 value);
   event RewardRuleAdded(address rewardProgramID, string ruleDID);
   event RewardRuleRemoved(address rewardProgramID, string ruleDID);
   event RewardeeRegistered(
@@ -47,6 +48,7 @@ contract RewardManager is Ownable, Versionable, Safe {
   address internal constant ZERO_ADDRESS = address(0);
   bytes4 internal constant EIP1271_MAGIC_VALUE = 0x20c13b0b;
   bytes4 internal constant SWAP_OWNER = 0xe318b52b; //swapOwner(address,address,address)
+  bytes4 internal constant TRANSFER = 0xa9059cbb; //transfer(address,uint256)
   string internal constant REWARD_PREFIX = "safe.rewards.cardstack";
   uint256 internal _nonce;
 
@@ -197,7 +199,7 @@ contract RewardManager is Ownable, Versionable, Safe {
     execTransaction(
       msg.sender,
       0,
-      getTransferRewardSafeData(msg.sender, newOwner),
+      abi.encodeWithSelector(SWAP_OWNER, address(this), oldOwner, newOwner),
       safeTxGas,
       baseGas,
       gasPrice,
@@ -209,6 +211,34 @@ contract RewardManager is Ownable, Versionable, Safe {
     emit RewardSafeTransferred(msg.sender, oldOwner, newOwner);
   }
 
+  function withdrawFromRewardSafe(
+    address token,
+    uint256 value,
+    uint256 safeTxGas,
+    uint256 baseGas,
+    uint256 gasPrice,
+    address gasToken,
+    bytes calldata signature
+  ) external {
+    address rewardSafeOwner = getRewardSafeOwner(msg.sender);
+    bytes memory conSignature = contractSignature(msg.sender, rewardSafeOwner);
+    signatures[keccak256(conSignature)] = true;
+
+    execTransaction(
+      token,
+      0,
+      abi.encodeWithSelector(TRANSFER, rewardSafeOwner, value),
+      safeTxGas,
+      baseGas,
+      gasPrice,
+      gasToken,
+      signature,
+      msg.sender
+    );
+    signatures[keccak256(conSignature)] = false;
+    emit RewardSafeWithdrawal(msg.sender, token, value);
+  }
+
   function getRewardSafeOwner(address payable rewardSafe)
     public
     view
@@ -217,15 +247,6 @@ contract RewardManager is Ownable, Versionable, Safe {
     address[] memory owners = GnosisSafe(rewardSafe).getOwners();
     require(owners.length == 2, "unexpected number of owners for reward safe");
     return owners[0] == address(this) ? owners[1] : owners[0];
-  }
-
-  function getTransferRewardSafeData(
-    address payable rewardSafe,
-    address newOwner
-  ) public view returns (bytes memory) {
-    address oldOwner = getRewardSafeOwner(rewardSafe);
-    return
-      abi.encodeWithSelector(SWAP_OWNER, address(this), oldOwner, newOwner);
   }
 
   function isRewardProgram(address rewardProgramID) public view returns (bool) {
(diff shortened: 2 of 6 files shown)


contracts/RewardManager.sol
```diff
@@ -224,6 +254,14 @@ contract RewardManager is Ownable, Versionable, Safe {
       abi.encodeWithSelector(SWAP_OWNER, address(this), oldOwner, newOwner);
   }
 
+  function getWithdrawFromRewardSafeData(
```
tintinthong, line 257: @alex-cardstack I think you can just pass the data in from offchain. Make a `data` param for this function. Mainly because I think the data for a transfer function is clear for ERC677 or ERC20 is unambiguous. Saves contract space. `getTransferRewardSafeData` saves client from knowing what the contract is. Using that logic, can you perhaps replace [this](https://github.com/cardstack/card-pay-protocol/blob/f145e7a4d1277e388ff71cfb1e1199e23c5b9836/test/utils/helper.js#L1298) with a smart contract call.

test/utils/helper.js
```diff
@@ -1380,6 +1380,98 @@ const transferRewardSafe = async function (
...
+    rewardSafe,
+    rewardManager
+  );
+
+  let safeTxData = {
+    to: rewardManager.address,
+    data: payload,
+    operation: 0,
+    txGasEstimate: 0,
+    gasPrice: 0,
```
tintinthong, line 1462: Perhaps, try adding `gasPrice`.  Since there are funds transferred we can collect gas -- look at `claimReward`. I don't do this for transferring ownership of safe because there may not be funds in the safe when it is transferred.

contracts/RewardManager.sol
tintinthong: Think this would save us contract space.

tintinthong:
Seems the ci is failing cos of a function `checkGnosisExecution`. Perhaps, you can fix it to handle the error when `payment` field is not there.

test/utils/helper.js
tintinthong: You can check the transfer by replicating this  [test](https://github.com/cardstack/card-pay-protocol/blob/13dbac1f5149004086818f01843d344bfc94c910/test/RewardPool-test.js#L672)

contracts/RewardManager.sol
tintinthong: Would like to hear your opinon on this? Becos in truth, we can also remove `getTransferRewardSafeData` honestly

tintinthong: @habdelra tagging you here because this has been a pattern for `transfer` in prepaid card manager also. I was wondering if there was an actual reason why `getTransferCardData` is written on chain when in truth the client can figure that out offchain and generate the data offchain.

habdelra: I think this is a relic from the original prepaidcard manager that Tim’s team worked on. If we aren’t calling this onchain anymore, then yeah we can remove it.

tintinthong: @alex-cardstack then can you get rid of `getTrasferRewardSafeData` and just make everything offchain. I just checked in the linked list implementation, so it is as expected that `prevOwner` can remain as the contract address and the `oldOwner` and `newOwner` are the only thing changing.

test/utils/helper.js
habdelra: @tintinthong i don't think it's worth while to change this from 0--that is all out of the box gnosis safe contract logic. It is the GnosisSafe.sol contract that actually deducts the gas fee out of the safe. That is not code we have control over--moreover gnosis has already written tests for it. I think it is only worthwhile to create test around gas price when we are actually doing something with it (like the prepaid card manager's gas policies). I would say leave this at 0 to make our lives simpler and rest assured on the fg\act that gnosis is testing their own contract to be able to deduct gas fees from the safes appropriately.

tintinthong: ok that is fine by me. I feel for me the value in doing this test is communication to someone who hasn't used gnosis stuff before. Honestly, when I first used these I would always wonder "why there is no gas fees paid out or why the gasPrice was 0". But if we just want our test to focus on the business logic and not be obscured by these gas payments -- I can see the benefit too.

contracts/RewardManager.sol
habdelra: @tintinthong I think we are both too laszer focused on the prepaidCardManager logic--`getWithdrawFromRewardSafeData` is actually used on chain--it's in the execTransaction herehttps://github.com/cardstack/card-pay-protocol/pull/131/files#diff-d99035ba0a6829c3a3a5f7ee88c5df2590f4c95c550100a03b81186e0e1f512aR226. But yes, we do have some left over stuff that in the PrepaidCardManager is no longer in the `execTransaction` after the `send()` that we did on it awhile back.

tintinthong: what do you mean? my proposal is to just remove the onchain function that calls for data such as `getWithdrawFromRewardSafeData` and `getTransferRewardSafeData` and pass the data in as an argument to the main function, such as `withdrawFromSafe` and `transferRewardSafe`. We will not touch the prepaid card manager bcos that is critical path.

alex-cardstack: @tintinthong had a call with @habdelra this morning, and he thinks you and he were thinking of the prepaid card manager send function, however in this case it does need to be on-chain because otherwise the client can send any transaction to the token that the contract will blindly sign - it being on-chain here is what verifies the contract owner and EOA owner actually agree on the same action

habdelra: @tintinthong oh my misunderstanding--I was the one with PrepaidCardManager on the brain then. I see what you mean. 

@alex-cardstack in this case the `withdrawFromRewardSafe` itself would have this data (the data that is returned from `getWithdrawFromRewardSafeData` as derivable with web3 or ethers) passed in as an additional param from what you already have,

tintinthong: Actually @habdelra I take everything back. @alex-cardstack I think is right.

The client can piggy back the `withdrawTransferFunction` to send whatever data he wants other than TRANSFER. So this thing has to be onchain.

tintinthong: Sorry for the confusion then. It seems that `getWithdrawFromRewardSafeData` and `getTransferRewardSafeData` all have to be functions on-chain.

contracts/RewardManager.sol
```diff
@@ -205,6 +207,34 @@ contract RewardManager is Ownable, Versionable, Safe {
...
+    uint256 baseGas,
+    uint256 gasPrice,
+    address gasToken,
+    bytes calldata signature
+  ) external {
+    address rewardSafeOwner = getRewardSafeOwner(msg.sender);
+    bytes memory conSignature = contractSignature(msg.sender, rewardSafeOwner);
+    signatures[keccak256(conSignature)] = true;
+
+    execTransaction(
```
tintinthong, line 223: @alex-cardstack but the safe will not trigger this right now if im not mistake because there is no signature allowing thru to the token contract.

test/utils/helper.js
alex-cardstack: > why there is no gas fees paid out or why the gasPrice was 0"

Heh yeah this was me. Also I had thought that changing the gas price / adding the estimate was the first thing that made it work at all, but I might be wrong there, I'm continuing to dig

contracts/RewardManager.sol
alex-cardstack: @tintinthong I am having problems getting this part to work - can you explain what you mean "there is no signature allowing thru to the token contract"?

Is that no the `signature` param to this function?

tintinthong: Sure. All safe execTransactions will go to the `isValidSignature` callback; this is the case for the execTransaction within this contract. I will comment at the code

alexspeller: OK nice, I will try that, thank you

contracts/RewardManager.sol
alexspeller: However, neither of these need to be public view functions at all - they are only used in a single place, I think it would save gas to just inline both `getWithdrawFromRewardSafeData` and `getTransferRewardSafeData`, they can both be a single line in the calling functions

contracts/RewardManager.sol
alex-cardstack: Thanks, that was the push in the right direction I needed. I have a further comment on my implementation there

contracts/RewardManager.sol
```diff
@@ -219,15 +249,6 @@ contract RewardManager is Ownable, Versionable, Safe {
     return owners[0] == address(this) ? owners[1] : owners[0];
   }
 
-  function getTransferRewardSafeData(
```
alex-cardstack, line 222: I inlined this function and saved some bytes, also some execution gas I think:



Before inline  getWithdrawFromRewardSafeData:
```
╰─➤ yarn test:size | grep RewardManager
 |  contracts/RewardManager.sol:RewardManager                                                            ·     14.623  │
```
After inline  getWithdrawFromRewardSafeData:
```
╰─➤ yarn test:size | grep RewardManager
 |  contracts/RewardManager.sol:RewardManager                                                            ·     14.598  │
```
After also inline getTransferRewardSafeData:
```
╰─➤ yarn test:size | grep RewardManager
 |  contracts/RewardManager.sol:RewardManager                                                            ·     14.505  │
```

contracts/RewardManager.sol
```diff
@@ -319,13 +345,25 @@ contract RewardManager is Ownable, Versionable, Safe {
...
     return
-      ((_equalBytes(data, encodedTransactionData) &&
-        (to == msg.sender && signatures[keccak256(contractSignature)])) ||
-        (to == address(this)) ||
-        (eip1271Contracts.contains(to)))
-        ? EIP1271_MAGIC_VALUE
-        : bytes4(0);
+      (to == msg.sender) ||
+      TokenManager(ActionDispatcher(actionDispatcher).tokenManager())
+        .isValidToken(to);
```
alex-cardstack, line 366: This is the implementation to see if it is an allowed token, is this check correct?

alex-cardstack (commented):
Please let me know if this implementation looks better - if it is looking good, I have a few unhappy path tests I'd like to had

README.md
```diff
@@ -115,6 +115,10 @@ The `RewardManager` contract is responsible for creating the gnosis safes that a
 1. The `rewardee` or the owner of the prepaid card that executed the `registerRewardee` action.
```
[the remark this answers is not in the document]
tintinthong: You can move it out. I put it there mainly because I wanted to check if our "assumed" order worked -- and it did. Maybe there is a use-case where the order of owners would change and we'd need to robustify how we transfer owners, but can be handled in separate pr.

README.md
```diff
@@ -115,6 +115,10 @@ The `RewardManager` contract is responsible for creating the gnosis safes that a
...
 2. The `RewardManager` contract itself.
 
+Withdraw is only possible to the EOA owner of the reward safe, using the `withdrawFromRewardSafe` action with a wrapped gnosis signature, the inner signature being for the `token.transfer(to, value)` call, and the outer signature being for the `withdrawFromRewardSafe` function.
+
+Only approved tokens can be withdrawn from the reward safe.
```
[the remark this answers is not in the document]
tintinthong: The current scope is to only transfer tokens which are federated by the token manager.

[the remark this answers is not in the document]
tintinthong: ie daicpxd and cardcpxd

contracts/RewardManager.sol
tintinthong: kewl!

contracts/RewardManager.sol
tintinthong: Yup!

tintinthong: Actually (sorry about this). This has me thinking on weather we want to have the `withdrawal` function on the contract. If we move the token to `eip1271Contracts` or another OR path, we don't even have to write a function on the reward manager. Meaning that a reward safe can literally do anything allowed on the token contract. We actually just have to move `signatures[keccak256(contractSignature)]` out of the way.

tintinthong: But if we move `signatures[keccak256(contractSignature)]`  actually the issue with this is the protection from re-entry/replay attacks bcos I guess the principal is that we have to assume the token contracts are "untrusted".

tintinthong: This might mean that `eip1271Contracts` pattern COULD be insecure. Right now it only crosses path with `claim` function which is the only real state-changing function, but I think the function abides by principals like state "changes first and external contract call last"

tintinthong (approved):
I added a non-conclusive comment about how we might want to interact with the token contracts via `isValidSignature` callback. It's not a change suggestion. Other than that its good!

contracts/RewardManager.sol
```diff
@@ -319,13 +345,25 @@ contract RewardManager is Ownable, Versionable, Safe {
       msg.sender,
       rewardSafeOwner
     );
+
+    bool isAllowedGnosisExecution = _equalBytes(data, encodedTransactionData) &&
+      signatures[keccak256(contractSignature)] &&
+      isAllowedGnosisRecipient(to);
+
+    bytes4 result = isAllowedGnosisExecution ||
+      (to == address(this)) ||
```
tintinthong, line 354: @alex-cardstack it looks like previously I missed out something. each or condition should check `_equalBytes(data,encodedTransactionData)`. Maybe just a `require(_equalBytes(data, encodedTransactionData)`. This is because we have to validate that `to` param is correct.

contracts/RewardManager.sol
```diff
@@ -319,13 +345,25 @@ contract RewardManager is Ownable, Versionable, Safe {
...
+      eip1271Contracts.contains(to)
```
tintinthong, line 355: @alex-cardstack I missed something out here. Each OR condition should check `_equalBytes(data,encodedTransactionData)`. This is to ensure fidelity; so people don't lie about `to` param.  Think it would be as simple as just `require(_equalBytes(data,encodedTransactionData))`

alex-cardstack:
I added some unhappy path tests, I think this is good to merge now

PR #132, opened 2021-11-03
Restructure reward pool

IanCal:
Trying to remove the partial claims while transitioning the contract to a state that can support more use cases.

Key changes:

* Store which leaf nodes have been claimed as a boolean rather than a total (#cs-2250 do not support someone taking half of their reward today and half tomorrow)
* * The balance functions have been deleted because of this
* Support the case when someone wants to withdraw tokens but there aren't enough (#CS-2457)
* Leaf node now contains all of the data required for a claim
* You can check the validity of a leaf node without having to claim (#CS-2052) which means we can support giving people a provable status without needing nfts.
* A token type has been added, so we can have
* * Typical claims of tokens
* * NFT claims, and distinguish between different cases (minting vs just transferring)
* * Non-tokens, where we just encode data to represent some status/other thing
* Start & end valid ranges for claims (#cs-2492)

This covers a few things partly because it all required changing the data structure so they'd all conflict. I can extract this out into different PRs for clarity if the broad strokes seem sensible.

Questions/outstanding:

* ~~Cycles are still one for everything, having one per reward program is the next step. As it's not tied with the data structure change I wanted to split that out~~ Multiple cycles PR here https://github.com/cardstack/card-pay-protocol/pull/143
* Actual NFT transfers - there's a space for this to happen but no actual code

---

Tickets covered:
#cs-2250
#cs-2052
#cs-2492
#cs-2457


tintinthong:
> This is obviously a large breaking change requiring SDK & backend changes as well. If this is where we want to end up, do we want to plan a smoother transition in stages?

I would say go crazy. Change everything in a pr but don't merge yet. Because we don't want hassan to need to handle deploys that feature flag changes. There is an emulation of how merkle trees work here in `test/utils/cumulative-payment-tree` and `test/utils/merkle-tree`. So I'd rather change the behaviour here first before we move anything to tally or the sdk. Also, lets use the pattern of having a big pr and multiple prs that merge into it.

tintinthong:
> NOTE - I don't understand the types in solidity at the moment, and am aware this doesn't build - hopefully the flow I'm proposing makes sense despite mixing up calldata/memory and conversions between bytes/bytes32/etc

From what I understand, `calldata` is specified only as a parameter of an external function and is non-mutable. See [here](https://ethereum.stackexchange.com/questions/74442/when-should-i-use-calldata-and-when-should-i-use-memory). But the implications to the dev is little.

contracts/RewardPool.sol
```diff
@@ -92,13 +92,39 @@ contract RewardPool is Initializable, Versionable, Ownable {
...
+    bytes calldata leaf,
+    bytes calldata proof
+  ) external view returns (bool) {
+    uint256 paymentCycleNumber = uint256(leaf[4]);
+    uint256 root = payeeRoots[paymentCycleNumber];
+    return proof.verify(root, keccak256(leaf));
+  }
+
+  function claimed(leaf) external view returns (bool) {
+    uint256 paymentCycleNumber = uint256(leaf[4]);
```
tintinthong: think you meant to use `abi.decode()` here.

contracts/RewardPool.sol
```diff
@@ -92,13 +92,39 @@ contract RewardPool is Initializable, Versionable, Ownable {
...
   ) external returns (bool) {
-    require(amount > 0, "Cannot claim non-positive amount");
+
+    uint256 rewardProgramID = uint256(leaf[0]);
+    uint256 payableToken = uint256(leaf[1]);
+    address payee = address(leaf[2]);
+    uint256 amount = uint256(leaf[3]);
+    uint256 paymentCycleNumber = uint256(leaf[4]);
+
+    require(msg.sender == payee, "Can only be claimed by payee");
```
tintinthong, line 126: I love this simplification into 3 requires.

contracts/RewardPool.sol
```diff
@@ -92,13 +92,39 @@ contract RewardPool is Initializable, Versionable, Ownable {
...
   function claim(
-    address rewardProgramID,
-    address payableToken,
-    uint256 amount,
+    bytes calldata leaf,
     bytes calldata proof
   ) external returns (bool) {
-    require(amount > 0, "Cannot claim non-positive amount");
+
+    uint256 rewardProgramID = uint256(leaf[0]);
```
tintinthong: why is rewardProgramID a `uint256`. I previously used the id as an ethereum address.

contracts/RewardPool.sol
```diff
@@ -92,13 +92,39 @@ contract RewardPool is Initializable, Versionable, Ownable {
...
+    uint256 payableToken = uint256(leaf[1]);
```
tintinthong: payableToken is definitely an address

contracts/RewardPool.sol
```diff
@@ -92,13 +92,39 @@ contract RewardPool is Initializable, Versionable, Ownable {
     return true;
   }
 
+  function valid(
+    bytes calldata leaf,
+    bytes calldata proof
```
tintinthong: im assuming here that proof does not have meta appended to it?

tintinthong: I notice you combined everything into leaf. So even things that were not in the original leaf such as `paymentCycle` was just included into leaf.

tintinthong:
> If the leaf node stores the amount to claim, and we know if it's been claimed (claimed function added) then I don't think we need the balance checking functions

Agreed. We do however, still need balance checking functions offchain.  So offchain balance function will need to check for weather a proof is claimed.

contracts/RewardPool.sol
IanCal: Yeah, my idea was to have two things, the leaf node and the proofs. Leaf is now kind of what was the meta & the calling arguments, and in my mind should contain everything needed to verify it's valid. 

Requires a change elsewhere obviously.

tintinthong:
> Leaf nodes can contain any extra data we need, and still be verifiable. The functions only care about the start of the structure

Think we would want to lock down on this because the merkle tree will hash the leaf. Mainly because of the `abi.encode` and `abi.decode`, which requires you to specify data types of the encoding. 

> This doesn't support NFT claiming, but with the leaf nodes being more flexible perhaps that's a non-breaking change to update

I think we need to think about this soon and change the leaf structure accordingly. I don't think we have or should allow that degree of freedom of extra bytes so we can make our mind what the leaf is.

contracts/RewardPool.sol
IanCal: Looks like we'd need to swap out a packed encoding for a non-packed one as we can't do abi.decodepacked (doesn't exist).
If that's the case, and it has to be a bytes32 array, then because addresses are 20 bytes we'd have lots of space in the "address" bytes32 to store info about the contract (such as whether it's an nft or not).

Clear cost of increased size of data & the hashing though.

tintinthong:
All in all. I think this is a FANTASTIC refactor of the reward pool contract. Tho I think the next step would be to finalise how nft fits into this if that is still a requirement. Because this change will propagate a lot of changes everywhere already so might as well kill all birds with one stone.

contracts/RewardPool.sol
```diff
@@ -220,49 +205,10 @@ contract RewardPool is Initializable, Versionable, Ownable {
...
-      return
-        cumulativeAmount.sub(
-          claims[paymentCycleNumber][rewardProgramID][payableToken][_address]
-        );
-    } else {
-      return 0;
-    }
-  }
-
   function splitIntoBytes32(bytes memory byteArray, uint256 numBytes32)
```
IanCal, line 208: remove this again by swapping for abi.decode in those places

IanCal:
This currently has an issue when two NFTs are claimable in a single reward cycle, as the claimed check would be the same for both. I'll sketch out swapping the claimed storage for a hash of the leaf node.

I've tried combining these using a more nested approach, but there are a few sticking points to do with ordering of the necessary fields.

test/utils/cumulative-payment-tree.js
```diff
@@ -22,45 +22,8 @@ const _ = require("lodash");
...
-    // find a better way to check this
-    if (!leaf) {
-      return bufferToHex(zeros(32));
-    }
-    return this.getHexProof(leaf, [
-      paymentCycle,
-      this.amountForPayee(rewardProgramID, payee, token),
-    ]);
+    super(paymentList);
+    this.paymentNodes = paymentList;
```
IanCal, line 26: This is so thin now the tree can almost be removed. Will refactor this a little to move the custom payment part into this and rename it.

contracts/RewardPool.sol
```diff
@@ -92,65 +93,112 @@ contract RewardPool is Initializable, Versionable, Ownable {
...
   function claim(
-    address rewardProgramID,
-    address payableToken,
-    uint256 amount,
-    bytes calldata proof
-  ) external returns (bool) {
-    require(amount > 0, "Cannot claim non-positive amount");
+    bytes calldata leaf,
+    bytes32[] calldata proof,
+    bool partialClaimAllowed
```
tintinthong, line 155: do we really want to maintain `partialClaimAllowed`

IanCal: Probably a better name required here, this is for the requirement of "what if there's not enough in the pool for the full claim, do you still want to just get as much as you can".

Could maybe be two clear claim functions?

contracts/RewardPool.sol
```diff
@@ -92,65 +93,112 @@ contract RewardPool is Initializable, Versionable, Ownable {
     return true;
   }
 
+  function valid(
+    bytes memory leaf,
+    bytes32[] memory proof
+  ) public view returns (bool) {
+        (
+      address rewardProgramID,           // solhint-disable-line no-unused-vars
```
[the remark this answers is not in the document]
alex-cardstack: actually I think it might be better if only for slightly reduce gas consumption not assigning those vars, but maybe the compiler optimises it out anyway

test/RewardPool-test.js
```diff
@@ -372,6 +406,11 @@ contract("RewardPool", function (accounts) {
           paymentCycle
         );
         root = merkleTree.getHexRoot();
+        rewardPool.submitPayeeMerkleRoot.estimateGas(root, { from: tally }).then(
+          function(gas) {
```
alex-cardstack, line 410: All this lint stuff is [prettier](https://prettier.io/), ideally your editor should run on save but not sure what editor you're using or how much of a yak shave it is to set that up

test/RewardPool-test.js
```diff
@@ -571,7 +610,7 @@ contract("RewardPool", function (accounts) {
       });
     });
 
-    describe("claim", function () {
+    describe.only("claim", function () {
```
alex-cardstack, line 613: `.only` left in

alex-cardstack (commented):
Looking good, I guess if we go the safe route for reward pool storage there will be some changes here but what's here makes sense

test/RewardPool-test.js
IanCal: Perfect, wanted this but didn't want to run a different thing to everyone else. Is easy to setup on save, will do that.

contracts/RewardPool.sol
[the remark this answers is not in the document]
IanCal: Yeah, had tried to mark them as unused in other ways but hadn't found one that worked, this does thanks! I initially tried adding in the commented variables to make it less likely something pulls the wrong part of the data, but the formatter pulls them out into weird places so have gone with the first format.

contracts/RewardPool.sol
```diff
@@ -92,55 +93,56 @@ contract RewardPool is Initializable, Versionable, Ownable {
...
-        msg.sender,
-        rewardProgramID
-      ),
-      "can only withdraw for safe registered on reward program"
+  function valid(bytes memory leaf, bytes32[] memory proof)
+    public
+    view
+    returns (bool)
+  {
+    (, uint256 paymentCycleNumber, , , , , ) = abi.decode(
```
alex-cardstack, line 101: Sorry to be a pain, but now that I see this I hate it even more, it feels like a big footgun. If the comments weren't pulled out by the annoying formatter then it would be better but actually still pretty error prone.

I think it would be really nice to have a single source of truth for "how to extract the `paymentCycleNumber` from leaf bytes, so a function `extractPaymentCycleNumber(bytes memory leaf) returns (uint256)` etc

Adding a bunch of getters like this in this contract will bloat bytes and not be reusable. I think that this is a great use case for a library with external view functions, which is deployed at it's own address on the blockchain and therefore doesn't count towards this contract's bytesize limit.

It's a bit nitpicky I know, but I feel having a SSOT for decoding this that can be used here and in other contracts will be a prudent way to do this and less error-prone as we use it more often.

contracts/RewardPool.sol
```diff
@@ -150,7 +152,71 @@ contract RewardPool is Initializable, Versionable, Ownable {
...
+  }
+
+  function claim(
+    bytes calldata leaf,
+    bytes32[] calldata proof,
+    bool acceptPartialClaim
+  ) external nonReentrant returns (bool) {
+    (
+      address rewardProgramID,
+      ,
```
alex-cardstack, line 164: Yeah this is worse to review / read IMO, sorry, I think library or previous approach if the library is too yak shavey.

alex-cardstack (changes requested):

IanCal:
@alex-cardstack I've rolled back to the previous way of pulling out the data for the leaf nodes

* Structs would be nice but the abi decoder in solidity 0.5 doesn't support nicely extracting into them, and the v2 encoder was not only experimental at that point but there are comments that there were some serious issues before 0.6
* Returning structs from a function also didn't work with a similar abi encode/decode error. Unsure if that's the older solidity or a misunderstanding on my side.

You may be right about using another contract, but I think it would be clearer as a change if it's on top of this as this already merges a bunch of changes together.

alex-cardstack (approved):

PR #133, opened 2021-11-08
Move away from create2 for reward safes

alex-cardstack:
This allows transferring a reward safe to another user, then letting
the original user re-register for the reward progresm

This requires storing and updating some state

Note that this change breaks storage layout so needs a new contract deploy instead of a proxy upgrade,
let me know if we need to do an upgrade for this and I'll refactor the storage ordering


habdelra:
it should be fine to break upgradablility--we still have no clients that depend on this contract address yet

habdelra (approved):
looks good to me

habdelra:
@alex-cardstack I just merged Justin's PR which looks like it created a conflict here

PR #139, opened 2021-11-10
extract unclaimed reward tokens by reward program admin

tintinthong:
- extracts to any safe

### DIFF
--- a/contracts/RewardPool.sol
+++ b/contracts/RewardPool.sol
@@ -171,7 +171,9 @@ contract RewardPool is Initializable, Versionable, Ownable {
     require(
       GnosisSafe(msg.sender).getOwners()[1] == rewardProgramAdmin,
       "owner of safe is not reward program admin"
-    ); // we recover to generic safe now. due attention in safe manager refactor
+    );
+    // we recover to generic safe now. due attention in safe manager refactor
+    // also this assumes an order in owner array
     require(
       rewardBalance[rewardProgramID][token] >= amount,
       "not enough tokens to withdraw"
--- a/contracts/RewardPool.sol
+++ b/contracts/RewardPool.sol
@@ -13,6 +13,7 @@ import "./RewardManager.sol";
 import "./TokenManager.sol";
 import "./VersionManager.sol";
 
+
 contract RewardPool is Initializable, Versionable, Ownable {
   using SafeMath for uint256;
   using MerkleProof for bytes32[];
@@ -36,6 +37,12 @@ contract RewardPool is Initializable, Versionable, Ownable {
     address sender,
     address tokenAddress,
     uint256 amount
+  );
+  event RewardTokensExtracted(
+    address rewardProgramID,
+    address token,
+    uint256 amount,
+    address rewardProgramAdmin
   );
 
   address internal constant ZERO_ADDRESS = address(0);
@@ -151,6 +158,30 @@ contract RewardPool is Initializable, Versionable, Ownable {
       amount
     );
     return true;
+  }
+
+  function extractTokens(
+    address rewardProgramID,
+    address token,
+    uint256 amount
+  ) external returns (bool) {
+    address rewardProgramAdmin = RewardManager(rewardManager).getRewardProgramAdmin(
+      rewardProgramID
+    );
+    require(
+      GnosisSafe(msg.sender).getOwners()[1] == rewardProgramAdmin,
+      "owner of safe is not reward program admin"
+    );
+    rewardBalance[rewardProgramID][token] = rewardBalance[rewardProgramID][
+      token
+    ].sub(amount);
+    IERC677(token).transfer(msg.sender, amount);
+    emit RewardTokensExtracted(
+      rewardProgramID,
+      token,
+      amount,
+      rewardProgramAdmin
+    );
   }
 
   function onTokenTransfer(
--- a/contracts/RewardPool.sol
+++ b/contracts/RewardPool.sol
@@ -165,12 +165,16 @@ contract RewardPool is Initializable, Versionable, Ownable {
     address token,
     uint256 amount
   ) external returns (bool) {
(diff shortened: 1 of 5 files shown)


contracts/RewardPool.sol
```diff
@@ -153,6 +160,30 @@ contract RewardPool is Initializable, Versionable, Ownable {
...
+    uint256 amount
+  ) external returns (bool) {
+    address rewardProgramAdmin = RewardManager(rewardManager).getRewardProgramAdmin(
+      rewardProgramID
+    );
+    require(
+      GnosisSafe(msg.sender).getOwners()[1] == rewardProgramAdmin,
+      "owner of safe is not reward program admin"
+    );
+    rewardBalance[rewardProgramID][token] = rewardBalance[rewardProgramID][
```
alex-cardstack, line 175: Maybe it's checked earlier but should balance be checked here?

```suggestion
    require(rewardBalance[rewardProgramID][token] >= amount, "Not enough tokens to withdraw")
    rewardBalance[rewardProgramID][token] = rewardBalance[rewardProgramID][
```

test/RewardPool-test.js
```diff
@@ -1730,6 +1731,88 @@ contract("RewardPool", function (accounts) {
         );
       });
     });
+
+    describe.only("extractRewardTokens", function () {
```
alex-cardstack, line 1735:
```suggestion
    describe("extractRewardTokens", function () {
```

test/RewardPool-test.js
```diff
@@ -1730,6 +1731,88 @@ contract("RewardPool", function (accounts) {
...
+        let rewardPoolBalance = await rewardPool.rewardBalance(
+          rewardProgramID,
+          cardcpxdToken.address
+        );
+        assert(
+          rewardPoolBalance.eq(toTokenUnit(50).sub(toTokenUnit(10))),
+          "reward pool balance is correct"
+        );
+      });
+    });
```
alex-cardstack, line 1815: Maybe worth adding some unhappy path cases like trying to withdraw without enough balance maybe

alex-cardstack (changes requested):

tintinthong:
Im going to rename everything to `recoverRewardTokens` rather than `extract`. Extract just sounds wrong and unnecessarily surgical.

tintinthong:
So @alex-cardstack I made a fix. I would appreciate your input on `setupProtocol` function. Currently it is only used in the reward pool test. But I created it witth the intention of making a god fixture-making function as a `before` in each test file. 

For example, reward pool the structure wud be like this 

- `before() setupProtocol` 
- `describe` > `beforeEach` > `RewardPool.new() & RewardPool.setup()` 

The implementation of `setupProtocol` is incomplete, but I wonder if it makes sense to actually do this in every contract test file and just destructure out the things that you need for your test. It would be a great to have if this `setupProtocol` function was intelligent enough to setup only the required dependecy thru some resolution logic. For example, if I require rewardPool, then rewardManger, tokenManager, versionManager, tally address are all setup for me. Wonder if there's a library to handle such a thing. 

hassan ahd the idea of making fixtures for different protocol classes -- like prepaid card & revenue (payment) / rewards.

alex-cardstack:
> So @alex-cardstack I made a fix in the test paying reward pool with a prepaid card this was screwed up by the setupProtocol. I would appreciate your input on setupProtocol function. Currently it is only used in the reward pool test. But I created it witth the intention of making a god fixture-making function as a before in each test file.

> For example, reward pool the structure wud be like this

> before() setupProtocol
> describe > beforeEach > RewardPool.new() & RewardPool.setup()

This makes sense. I did have some issues with the before being run before all tests not beforeEach - I worked around it, it wasn't terrible.

I do like the idea of it being easy to have the whole ecosystem set up easily for every test. It makes it much easier to test things.

> The implementation of setupProtocol is incomplete, but I wonder if it makes sense to actually do this in every contract test file and just destructure out the things that you need for your test. It would be a great to have if this setupProtocol function was intelligent enough to setup only the required dependecy thru some resolution logic. For example, if I require rewardPool, then rewardManger, tokenManager, versionManager, tally address are all setup for me. Wonder if there's a library to handle such a thing.

I can't think of one off the top of my head, maybe we can just do it with lazy accessors.

So we could have e.g. this API:

```js
import { getRewardManagerInstance } from "../setup";
let rewardManager = await getRewardManagerInstance()
```
And then implement it with:
```js
const rewardManagerInstance, paymentManagerInstance;

export async function getRewardManagerInstance() {
  if (rewardManagerInstance) { return rewardManagerInstance; }
  ...
  let paymentManagerInstance = await getPaymentManagerInstance();
  rewardManagerInstance = deploy(...);
  await rewardManagerInstance.setup(await getPaymentManagerInstance())
}

export function getPaymentManagerInstance() {
  if (paymentManagerInstance) { return paymentManagerInstance; }
  ...

}
```

This pattern would be composable - caching the deployed contracts so that they could be used individually. Essentially you get dependency ordering "for free" by just using the function calls and memoizing the results.

> hassan ahd the idea of making fixtures for different protocol classes -- like prepaid card & revenue (payment) / rewards.

However we setup the tests, I think it's basically a tradeoff of test running speed vs fidelity vs ease of maintenance and I'm not sure what's best at the moment.

Maybe just setting up the whole protocol at the beginning is the simplest way to make it realistic, I'm not sure

contracts/RewardPool.sol
```diff
@@ -153,6 +160,46 @@ contract RewardPool is Initializable, Versionable, Ownable {
...
+      rewardProgramID,
+      token,
+      amount,
+      rewardProgramAdmin
+    );
+  }
+
+  // lazy implementation of getting eoa owner of safe that has 1 or 2 owners
+  // think this is a use-case to handle during safe manager refactor
+  function _getEOAOwner(address safe) internal returns (address) {
```
alex-cardstack, line 193: Yeah I think this definitely is not guaranteed to be right, I agree that the refactor will solve

alex-cardstack (approved):

PR #143, opened 2021-11-17
Split out roots into one per payment cycle and per reward program

IanCal:
#CS-2250

Each reward program should have its own cycles. This allows for pulling an erroneous root with only touching one reward program (not implemented but the data change here would be required).

This pushes more work into tally - mainly here specifying the payment cycle upfront rather than submitting and having it autoincrement. That change would largely be required anyway with the other PR due to the change in data structure - this makes that cleaner.

There may be other feature improve use (e.g. check if a root has been submitted yet for a payment cycle in a reward program) but I thought it best to keep it shorter.

Notes:

* Claim functions are untouched - since the leaf node already includes the program ID and cycle number, the change there is transparent to the end user.
* Tests could do with refactoring at some point, there's a lot of repeated constructions of merkle trees.

Questions:

* Event change here, what information do we really need out? Do we still need block numbers for the range?


tintinthong:
Sorry for late review. 

> Event change here, what information do we really need out? Do we still need block numbers for the range?

It is not needed, it actually comes for free in the subgraph. for example, [here](https://github.com/cardstack/cardstack/blob/ceef58cf774f7ead58aae4cf689232239ac8b96a/packages/cardpay-subgraph/src/mappings/reward-pool.ts#L7). Tho we have not built a subgraph for merkle root submission.

contracts/RewardPool.sol
```diff
@@ -111,7 +110,7 @@ contract RewardPool is Initializable, Versionable, Ownable, ReentrancyGuard {
         (address, uint256, uint256, uint256, uint256, address, bytes)
       );
     if (block.number >= startBlock && block.number < endBlock) {
```
tintinthong, line 112: Hmmmm, I might have missed this. how does tally know of both the startBlock and endBlock of a payment cycle?

It could be that enforcing tally to know this information may add to some complexity.

tintinthong:
@IanCal it might be an idea to add a public mapping or a public function that queries the currentPaymentCycle of a rewardProgram.

test/RewardPool-test.js
```diff
@@ -240,144 +243,103 @@ contract("RewardPool", function (accounts) {
...
-          currentBlockNumber,
-          "the payment cycle end block is correct"
+          merkleRootSubmissionEvent.args.rewardProgramID.toLowerCase(),
+          rewardProgramID.toLowerCase(),
+          "the reward program ID is correct"
         );
       });
 
-      it("allows a new merkle root to be submitted in a block after the previous payment cycle has ended", async function () {
+      it("allows a multiple merkle roots for the same reward program to be submitted", async function () {
```
tintinthong, line 282: ^,

test/RewardPool-test.js
```diff
@@ -240,144 +243,103 @@ contract("RewardPool", function (accounts) {
...
-        );
       });
 
       it("does not allow non-tally to submit merkle root", async function () {
         let merkleTree = new PaymentTree(payments);
         let root = merkleTree.getHexRoot();
 
         await rewardPool
-          .submitPayeeMerkleRoot(root, { from: accounts[11] })
+          .submitPayeeMerkleRoot(rewardProgramID, 1, root, {
```
tintinthong, line 330: Yea. After looking at the test. It seems silly to rely on a paymentCycle specified by the caller to submit a merkle root. Shouldn't the smart contract know which paymentCycle you are wrong within its own state.

contracts/RewardPool.sol
```diff
@@ -80,15 +76,18 @@ contract RewardPool is Initializable, Versionable, Ownable, ReentrancyGuard {
...
 
-  function submitPayeeMerkleRoot(bytes32 payeeRoot)
-    external
-    onlyTally
-    returns (bool)
-  {
-    payeeRoots[numPaymentCycles] = payeeRoot;
+  function submitPayeeMerkleRoot(
+    address rewardProgramID,
+    uint256 paymentCycle,
```
tintinthong, line 81: I'd say remove `paymentCycle` here. And keep track of it inside the contract.

tintinthong (changes requested):

contracts/RewardPool.sol
tintinthong: So I was just pondering this. This may be a guard from double-writing merkle roots. The paymentCycle might act like a nonce.

contracts/RewardPool.sol
IanCal: Yeah with what I've done tally would need to know the range each payment cycle covers - I'll come back to that in another comment but the code here is actually the range of validity for the claim rather than anything to do with the payment cycle. I'll rename these variables as I think it's confusing now you ask about them.

e.g. You have reward of 500 airmiles but the claim expires in a year and after that point if you've not claimed them that's it.

Start block is "valid from" and end block is "valid to" for the claim.

test/RewardPool-test.js
IanCal: So my idea is that each payment cycle is a fixed number of blocks, so that each miner / etc knows the range it's supposed to be calculating the rewards for (or at least how frequently to run) and they don't necessarily have to be run in order.

> Like what if tally loses track of the payment cycle. Is there a deeper reasoning behind this??

If you know the payment cycles you want calculated, you can see if they've already been calculated and submitted (or should be able to, I've realised the variable is private). Tally can keep track of which ones it's already done, but that is essentially an optimisation.

If we're catching the merkerootsubmission event in subgraph we would also have a record there of which ones had been done already.

We could make it sequential again (at least for now, and adjust later when we want to bring in decentralisation), the overall interface is very similar. In tally, we get the current payment cycle just before writing the root anyway, so could add it back into the contract code.

contracts/RewardPool.sol
tintinthong: Right goddit. This is the expiry. Cool

test/RewardPool-test.js
tintinthong: > So my idea is that each payment cycle is a fixed number of blocks, so that each miner / etc knows the range it's supposed to be calculating the rewards for (or at least how frequently to run) and they don't necessarily have to be run in order.

Oh I see. Yes this is big change from previous implementation. Cool. **So computation can be out-of-sync with the write.** You can have cycle1, cycle3 already computed, with cycle2 still pending. **And writes don't have to be sequential.** But cycle3 can be written first, before cycle1. 

Correct?

> In tally, we get the current payment cycle just before writing the root anyway, so could add it back into the contract code.

So the choice here is between querying the payment cycle in the subgraph or on the contract. If we specify it in the contract, we will HAVE to assume sequentialism. So what I understand is that u intentionally don't put it in the contract, so tally can just query the last 4 root submissions, it might be (cycle1, ??, cycle3, cycle4). And then tally will decide if the computation for cycle2 is ready for write or if the computation for cycle5 is ready for write). So we can enforce sequentialism based on this decision by tally and information from the subgraph or we can just make it parallel. 

Correct?

IanCal: > Oh I see. Yes this is big change from previous implementation. Cool. So computation can be out-of-sync with the write. You can have cycle1, cycle3 already computed, with cycle2 still pending. And writes don't have to be sequential. But cycle3 can be written first, before cycle1.

Yes - that's right. While we're at this stage, we could make it sequential.

> So the choice here is between querying the payment cycle in the subgraph or on the contract. If we specify it in the contract, we will HAVE to assume sequentialism. So what I understand is that u intentionally don't put it in the contract, so tally can just query the last 4 root submissions, it might be (cycle1, ??, cycle3, cycle4). And then tally will decide if the computation for cycle2 is ready for write or if the computation for cycle5 is ready for write). So we can enforce sequentialism based on this decision by tally and information from the subgraph or we can just make it parallel.

> Correct?

Broadly, yes I think so.

I think this PR isn't complete actually - we need to either add things back in to support sequential work or also store the payment cycle length somewhere for each reward program.

IanCal: Or this could be in the rule blob...

tintinthong: > I think this PR isn't complete actually - we need to either add things back in to support sequential work or also store the payment cycle length somewhere for each reward program.

Right this PR implementation seems to make things clear in a generalised sense. 

My first thought (in the big picture) is that in terms of logic, we don't need to enforce that merkle root writes occur one after another like we did before. If the computing node has incentive to compute the rewards, they will compute the rewards. If the writing node (or consensus game) has incentive, the root to the missing payment cycles should be written in. So perhaps there's no need to add anything anymore?

But this is so far down-the-line. Is this pr included in the audit?

tintinthong (approved):

PR #144, opened 2021-11-23
Update rewards readme

tintinthong:
- just wanna kick this off
- there is certainly more detail that CAN be given, but do we want to surface them in this README?


IanCal (approved):
Looks good, thanks!

PR #145, opened 2021-11-23
Reward withdrawal and transfer 

alex-cardstack:
Use delegate implementation for reward safe withdrawal and transfer

### DIFF
--- a/contracts/RewardSafeDelegateImplementation.sol
+++ b/contracts/RewardSafeDelegateImplementation.sol
@@ -3,8 +3,7 @@ pragma solidity ^0.5.17;
 import "@openzeppelin/contract-upgradeable/contracts/token/ERC20/IERC20.sol";
 import "./RewardManager.sol";
 import "./ActionDispatcher.sol";
-
-import "hardhat/console.sol";
+import "@gnosis.pm/safe-contracts/contracts/GnosisSafe.sol";
 
 // This contract is used as an implementation for delegatecall usage of safe operations.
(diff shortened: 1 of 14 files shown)


habdelra (approved):
Awesome!

contracts/RewardSafeDelegateImplementation.sol
```diff
@@ -0,0 +1,75 @@
...
+
+import "@openzeppelin/contract-upgradeable/contracts/token/ERC20/IERC20.sol";
+import "./RewardManager.sol";
+import "./ActionDispatcher.sol";
+import "@gnosis.pm/safe-contracts/contracts/GnosisSafe.sol";
+
+// This contract is used as an implementation for delegatecall usage of safe operations.
+// It will never be called from it's own deployment address, whenever code here executes,
+// address(this) will be the gnosis safe, and if you call a function on another contract,
+// that other contract will see msg.sender as the safe address
```
tintinthong, line 11: Can we add a note here stating that the `msg.sender` on this contract is the sender of the transaction. In our case the relayer

tintinthong (approved):
I finally understand why a CALL can't work while a DELEGATECALL can work. And why a CALL could work for the reward pool `claim` function. It's mainly bcos of `address(this)`=`rewardSafeAddress`, which allows you to transfer the tokens.  

Cool very awesome and thanks for doing swapOwner as well.

alex-cardstack:
I cleaned this all up I think it should be ready for final review / merge now

tintinthong (approved):
LGTM

PR #147, opened 2021-11-24
Add script to prepare abis

tintinthong:
Didn't use codemod. But think this is sufficient

 This is meant to prep the `artifacts` (which are in json) such that it can be easily copied as typescript files into the monorepo

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,4 @@ artifacts
 build
 typechain
 dist
+abi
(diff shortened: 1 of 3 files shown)


tintinthong:
Still need to handle naming of `price-oracle.ts`

alex-cardstack:
I think ideally we'd use typechain for this - we're already generating really nice `.d.ts` files in the typechain folder for all the contracts.

That becomes easier once we integrate the protocol repo into the monorepo

alex-cardstack (approved):

PR #148, opened 2021-11-26
Rename start/end blocks to validFrom/To for clarity

IanCal:
After some confusion in another review about what the start & end referred to, these have been renamed to talk about when the claim is _valid_. Does not change any behaviour.


tintinthong (approved):
LGTM

PR #149, opened 2021-11-26
Publicly visible structure for which roots have been submitted for the rewards program

IanCal:
This makes the contract the queryable source of truth about which payment cycles have been submitted already. This becomes more important when we move towards fixed reward cycles & later decentralisation.


tintinthong (approved):
LGTM. I don't know how useful this is. But it will certainly give us the flexbility if we want to identify this mapping for within contracts (and a more straightforward way to call this mapping in the outside world).

IanCal:
Yeah this is more as a change that I think becomes more useful later on, but timing means it'd be good to get this in the review in case public/private makes a difference for security.

PR #150, opened 2021-11-26
Only allow roots to be submitted for registered reward programs

IanCal:
A simple check to ensure we don't submit merkle roots for reward programs that aren't registered, which should avoid a class of errors that would otherwise only become apparent after users try and make claims.


tintinthong (approved):
LGTM

PR #151, opened 2021-11-26
Improve reward safe events

alex-cardstack:
I noticed that the event emission for rewards safe was inconsistent.

Following some discussion, it's unclear if it's easy to have these events
on the safe itself via the delegate or not at the moment from a subgraph POV

I have moved the other event to the safe for now.

My reasoning:

If it's easy to make the subgraph handle this, I think this might be the better
place for the events to live.

If it's not, then we won't rely on it and can move it later, whereas with
the inconsistent location, we might come to rely on events that we later move

This at least makes it consistent and we can move later if we need to for
technical compromise reasons


tintinthong (approved):
LGTM

PR #156, opened 2021-12-01
Add leaf in RewardeeClaim event

tintinthong:
- Tally has to use the subgraph to track `RewardeeClaim`s. This is so Tally can update it's dbs such that it knows about `proofs` that are already claimed. 
- The leaf is an id we use to identify each proof in the dbs. We can filter by this id and hence mark the proof as claimed.

### DIFF
--- a/test/RewardManager-test.js
+++ b/test/RewardManager-test.js
@@ -1261,7 +1261,7 @@ contract("RewardManager", (accounts) => {
       );
     });
 
-    it.only("can transfer reward safe", async () => {
+    it("can transfer reward safe", async () => {
       let tx = await registerRewardee(
         prepaidCardManager,
         prepaidCard,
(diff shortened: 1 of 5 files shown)


test/RewardManager-test.js
```diff
@@ -1260,7 +1261,7 @@ contract("RewardManager", (accounts) => {
       );
     });
 
-    it("can transfer reward safe", async () => {
+    it.only("can transfer reward safe", async () => {
```
IanCal, line 1264: test still isolated, that'll be what's causing the coverage failure

test/RewardManager-test.js
```diff
@@ -1304,6 +1305,7 @@ contract("RewardManager", (accounts) => {
         oldOwner: prepaidCardOwner,
         newOwner: otherPrepaidCardOwner,
       });
+      console.log(params[0]);
```
IanCal, line 1308: leftover debug

test/RewardManager-test.js
tintinthong: Thx!

habdelra:
@tintinthong would you like to deploy a new RewardPool contract instance for this as well? (i think that will simplify your subgraph handling so you dont have to deal with a contract whose event changes shape partway through)

habdelra (approved):
👍

IanCal (approved):

PR #157, opened 2021-12-02
Fix reward safe transfer attack

tintinthong:
- Attack Scenario: prepaidCardOwner owns rewardSafe1, otherPrepaidCardOwner owns rewardSafe2. otherPrepaidCardOwner can attack prepaidCardOwner by transferring rewardSafe2 to rewardSafe1.
- **I don't know why. but the require is NOT WORKING in `willTransferRewardSafe`**

### DIFF
--- a/contracts/RewardManager.sol
+++ b/contracts/RewardManager.sol
@@ -175,6 +175,11 @@ contract RewardManager is Ownable, Versionable, Safe {
   function willTransferRewardSafe(address newOwner) external {
     address oldOwner = getRewardSafeOwner(msg.sender);
     address rewardProgramID = rewardProgramsForRewardSafes[msg.sender];
+    require(false,"why does this not revert the transaction");
+    require(
+      ownedRewardSafes[newOwner][rewardProgramID] == address(0),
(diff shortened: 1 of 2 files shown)


alex-cardstack:
@tintinthong it was the execution result - it was correctly reverting but the test was always passing anyway

tintinthong:
> @tintinthong it was the execution result - it was correctly reverting but the test was always passing anyway

Phew ok thanks. Yea it was failing silently. This is still different from other test which would typically give u a stack trace revert on the transaction. This occurs for DELEGATE_CALLs only -- so I guess we must treat errors differently with DELEGATE_CALLs. Think error messages might defer in the relayer and rpc node too.

tintinthong:
> @tintinthong it was the execution result - it was correctly reverting but the test was always passing anyway

phew ok thanks. Usually there is a transaction reverted stack trace. I think this occurs only for DELEGATE_CALL. I think perhaps this also mean we will see different errors from the rpc node (hence relayer).

alex-cardstack (approved):

PR #158, opened 2021-12-24
Upgrade gnosis safe to 1.3 and solidity version to 0.8.9

alex-cardstack:
Lots of dependency syncing needed here.

We're at the latest solidity version supported by hardhat, and upgrading minor solidity versions should be simple once hardhat catches up.

Things to confirm / think about:

- [x] abicoderv2 is the default for solidity 0.8. For some reason this gives an error `CompilerError: Stack too deep when compiling inline assembly: Variable dataEnd is 1 slot(s) too deep inside the stack.`. I had to pin to the old abicoderv1 using a pragma to get anything to compile.
- [x] SafeMath is gone - this is included by default in solidity
- [x] Every contract name from oz contracts ends with `Upgradeable` - this is just a naming scheme and confusingly doesn't really mean the contract is upgradeable. For example, EnumerableSetUpgradeable is _not_ upgradeable between different OZ versions
- [x] I added a hardhat task to check for compiler warnings and error on CI. There were a bunch of compiler warnings that I fixed, largely around not always returning a value in functions that return a success bool. I think I found an actual bug here though, in [the daiPrice function of ChainLinkFeedAdapter contract](https://github.com/cardstack/card-pay-protocol/blob/main/contracts/oracles/ChainlinkFeedAdapter.sol#L123-L124). If `tokenUsdFeed == daiUsdFeed`, `price` and `updatedAt` return values are assigned on L123-124. However they are not returned, and are _always_ overwritten later [here](https://github.com/cardstack/card-pay-protocol/blob/main/contracts/oracles/ChainlinkFeedAdapter.sol#L135-L138). Would appreicate a particular look at b2368a8a5d6f91bce337da8f38f0a7cc4729adfa
- [x] We should probably not merge this until we're a lot more ready to migrate, in case of needing to deploy prod fixes based off main branch


habdelra:
@alex-cardstack good catch on the daiPrice function--yeah that `tokenUsdFeed == daiUsdFeed` was meant to be an early return from that function, we need an `else` block that surrounds the rest of the function since we are actually using named return values for the function

contracts/token/ERC677Token.sol
```diff
@@ -1,77 +1,77 @@
...
+    string memory _name,
+    string memory _symbol,
+    uint8 _decimals,
+    address owner
   ) public initializer {
-    _name = name;
-    _symbol = symbol;
-    _decimals = decimals;
-    initialize(minter);
+    // lots of initializers to override here, due to owner by default being msg sender
```
habdelra, line 25: There is no need really for this particular contract to be upgradeable--it's only ever used in our tests. If it makes your life easier, then let's just use a non-upgradeable ERC677

habdelra (approved):
This looks fantastic!

tintinthong (approved):
lgtm

IanCal:
This looks like it'll make it easier to work with other tools, slither works out of the box now and I easily got the forge compiler working (dapptools version not quite but not worth it with the forge project ongoing) so this will open up some really nice things for us going forward for safety.

PR #162, opened 2022-01-28
[CS-2908] Gnosis migration scripts

alex-cardstack:
* Added tested migration against sokol / xdai
* wrote migration scripts
* migration is in-place, no contract addresses change
* data is upgraded - current caveat is this needs _loads_ of trezor signing and
  might need batching to make it practical

TODO:

- [x] Final audit discovered one last struct field to migration - RevenuePool.RevenueBalance needs inner address set migration
- [x] solhint is barfing for some reason, need to workaround before linting passes


.openzeppelin/addresses-sokol.json
```diff
@@ -95,18 +95,6 @@
...
-    "contractName": "Market"
-  },
-  "Inventory": {
-    "proxy": "0xAd02eD619f234c4eb252b8b3636522fA24F7F6A8",
-    "contractName": "Inventory"
-  },
-  "LevelRegistrar": {
-    "proxy": "0x330e167945eb4EC0CB3a28D63533789aba62610B",
-    "contractName": "LevelRegistrar"
-  },
```
habdelra, line 109: if we do this, that triggers a new contract deploy for these contracts, which means new addresses, which means you need to also update the subgraph to accommodate these new addresses (this is what I was concerned about in discord). Failure to do so will result in our subgraph no longer indexing these contracts.

contracts/core/Ownable.sol
```diff
@@ -17,4 +17,7 @@ contract Ownable is OwnableUpgradeable {
       _transferOwnership(owner);
     }
   }
+
+  // add padding as storage layout changed in OZ contracts v4
+  uint256[1] private ____gap_Ownable;
```
habdelra, line 22: is it possible to actually use these with the latest hardhat upgrades lib from OZ?

.openzeppelin/addresses-sokol.json
habdelra: or wait--actually i remember now--these were the old commerce project's contracts. nevermind, these are not indexed by the subgraph

habdelra (approved):
looks good! 👍

.openzeppelin/addresses-sokol.json
alex-cardstack: @habdelra that makes sense, however I am removing this contract as it is no longer one of the contracts in the protocol - this artefact is a leftover

alex-cardstack:
Note: xdai tests pass locally, the failure on github actions seems to be due to rpc timeout errors

habdelra:
@alex-cardstack we have our own xdai rpc server if you need one. The url is available in the SDK’s constants module.

contracts/core/Ownable.sol
alex-cardstack: I have figured this out now, to enable it I had to:

1. Use `prepareUpgrade` to get the new storage layout metadata output in the json files
2. As part of the upgrade, fix that metadata to point to the newly deployed contract
3. Future upgrades work as normal.

This is tested by:

1. Start a local forked node `yarn start-forked-node:sokol`
2. Run the migration against the fork: `yarn migrate:fork:sokol`
3. Make a change to an arbitrary contract
4. Deploy to the forked node: `yarn deploy:fork:sokol`


Checked that the checks are still good by making a bad change:
image

Made a good change and deployed:

image


Verified working storage layout changes after deploy:
image

habdelra: that is SOOOO nice!

PR #165, opened 2022-02-24
Commit config change to new safe master copy

alex-cardstack:
Config change applied to sokol already, not xdai yet


lukemelia:
@alex-cardstack I think we can merge this, correct?

lukemelia:
Actually, I suppose this will need to be set in waypoint? When you say you applied this in Sokol, what do you mean exactly?

alex-cardstack:
@lukemelia so this was a FYI PR, I already applied the changes.

These .env params are used to set the on-chain config using `yarn deploy:config` from this repo - it calls the `setup` function on the smart contracts if any of the params have changed

PR #166, opened 2022-03-02
Add a reentrancy guard to delegate safe implementation

alex-cardstack:
Fixes CS-2882

Not exploitable due to gnosis signing required and whitelisting of tokens, but defence in depth


tintinthong (approved):
LGTM

PR #168, opened 2022-03-04
Refund when excess funds are accidentally transferred

alex-cardstack:
Audit feedback: Prevents EOAs losing tokens when accidentally transferring too many tokens

### DIFF
--- a/contracts/PrepaidCardManager.sol
+++ b/contracts/PrepaidCardManager.sol
@@ -534,12 +534,14 @@ contract PrepaidCardManager is Ownable, Versionable, Safe {
     }
 
     // refund the supplier any excess funds that they provided
-    if (
-      amountReceived > neededAmount &&
+    if (amountReceived > neededAmount) {
       // check to make sure ownerSafe address is a depot, so we can ensure it's
-      // a trusted contract
-      SupplierManager(supplierManager).safes(depot) != address(0)
-    ) {
+      // a trusted contract. Otherwise, revert to prevent caller losing funds
+      require(
+        SupplierManager(supplierManager).safes(depot) != address(0),
+        "Excessive funds sent for requested amounts"
+      );
+
       // the owner safe is a trusted contract (gnosis safe)
       IERC677(token).transfer(depot, amountReceived - neededAmount);
     }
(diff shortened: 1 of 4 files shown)


contracts/PrepaidCardManager.sol
```diff
@@ -534,12 +534,14 @@ contract PrepaidCardManager is Ownable, Versionable, Safe {
...
-      amountReceived > neededAmount &&
+    if (amountReceived > neededAmount) {
       // check to make sure ownerSafe address is a depot, so we can ensure it's
-      // a trusted contract
-      SupplierManager(supplierManager).safes(depot) != address(0)
-    ) {
+      // a trusted contract. Otherwise, revert to prevent caller losing funds
+      require(
+        SupplierManager(supplierManager).safes(depot) != address(0),
+        "Excessive funds sent for requested amounts"
```
tintinthong, line 542: The message seems not to address what is in the `require()`. Which is to check that the sender of funds is the depot so that an appropriate refund can be made.

tintinthong: That being said, shouldn't there be a `require()` that the sender of funds **is the depot** regardless of `amountReceived > neededAmount`.

tintinthong: I just read the audit finding. 

```
Exploit Scenario
A non-depot calling contract initiates a token transfer to a PrepaidCardManager contract,
invoking its createMultiplePrepaidCards() function and transferring more tokens
than necessary to top up the requested set of new cards. When the transaction is
complete, the extra tokens are not returned to the caller.
```

Doesn't this mean to remove the `require` or am I missing something.

tintinthong: > It might make sense to always require the sender of the funds is a supplier safe, I'm not sure why it was done this way in the first place, perhaps it should always check?

Id say to always check. For clarity, id put the require outside of the conditional `amountReceived > neededAmount`

tintinthong: Guess we should check if non-depots are creating prepaid cards??

habdelra: We actually do want non-depots to create prepaid cards. This logic needs to go.

alex-cardstack: @habdelra I removed the check for depots now, the logic is simpler it just always refunds overpayment

PR #170, opened 2022-03-09
Add comment noting isContract usage

alex-cardstack:
CS-2876

### DIFF
--- a/contracts/token/ERC677Token.sol
+++ b/contracts/token/ERC677Token.sol
@@ -49,6 +49,7 @@ contract ERC677Token is IERC677, ERC20PresetMinterPauserUpgradeable {
 
     emit Transfer(msg.sender, _to, _value, _data);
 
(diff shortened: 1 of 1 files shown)


contracts/token/ERC677Token.sol
```diff
@@ -49,6 +49,7 @@ contract ERC677Token is IERC677, ERC20PresetMinterPauserUpgradeable {
 
     emit Transfer(msg.sender, _to, _value, _data);
 
+    // Note: isContract() is not guaranteed to return an accurate value, never use it to provide an assurance of security.
```
jurgenwerk, line 52: I'm interested in knowing why is it not guaranteed?

jurgenwerk (approved):

contracts/token/ERC677Token.sol
alex-cardstack: There are some useful insights here: https://ethereum.stackexchange.com/questions/15641/how-does-a-contract-find-out-if-another-address-is-a-contract

PR #171, opened 2022-03-09
Add a sane limit to rewardProgramRegistrationFeeInSPEND

alex-cardstack:
Prevent setting an arbitrarily large fee in case of config error

CS-2874


jurgenwerk (approved):

tintinthong:
For a reward program 1000 SPEND = $10 seems like very small fee to be capped by.

tintinthong:
> @tintinthong trying to close this out, what do you think the limit should be? Is $1000 sensible? What's the highest registration fee you can imagine?

Given that this is configurable, I am happy with $1000

PR #172, opened 2022-03-11
Add zero value checks for contract setup functions and others

alex-cardstack:
CS-2872


contracts/PrepaidCardManager.sol
```diff
@@ -151,6 +159,8 @@ contract PrepaidCardManager is Ownable, Versionable, Safe {
     versionManager = _versionManager;
     Safe.setup(_gsMasterCopy, _gsProxyFactory);
     for (uint256 i = 0; i < _contractSigners.length; i++) {
+      require(_contractSigners[i] != address(0), "contractSigners not set");
```
jurgenwerk, line 162: I wanted to suggest `"contractSigner ${i} not set"` but now I see Solidity doesn't support easy string concatenation and it also costs more gas. Do you think it would be clearer to say "contractSigners should not contain empty addresses" or something similar?

contracts/RewardManager.sol
```diff
@@ -141,11 +157,16 @@ contract RewardManager is Ownable, Versionable, Safe {
...
+    require(
+      rewardProgramIDs.contains(rewardProgramID),
+      "invalid rewardProgramID"
+    );
     rule[rewardProgramID] = blob;
     emit RewardRuleAdded(rewardProgramID, blob);
   }
 
   function lockRewardProgram(address rewardProgramID) external onlyHandlers {
+    rewardProgramIDs.contains(rewardProgramID);
```
jurgenwerk, line 169: Can probably remove this line?

jurgenwerk (approved):
I left a couple of comments but otherwise this looks good!

contracts/PrepaidCardManager.sol
alex-cardstack: Yeah I like this wording better, correct that string interpolation in solidity is a pain, and for the sake of a revert message definitely a costly usage of gas

contracts/RewardManager.sol
alex-cardstack: Good catch, should have been a require, fixed

PR #174, opened 2022-03-14
Disable optimizations except special case

alex-cardstack:
[CS-2869](https://linear.app/cardstack/issue/CS-2869/solidity-compiler-optimizations-can-be-problematic)


tintinthong:
[CS-2869](https://linear.app/cardstack/issue/CS-2869/solidity-compiler-optimizations-can-be-problematic)?

tintinthong (approved):
This looks good!

PR #177, opened 2022-03-18
Safe transfer 

jurgenwerk:
Ticket: CS-2864

Replace `transfer` calls with `safeTransfer` so that if a transfer fails, the transaction will also fail.


alex-cardstack (approved):
Awesome, thanks!

PR #178, opened 2022-03-18
Introduce safeTransferAndCall

jurgenwerk:
Ticket: https://linear.app/cardstack/issue/CS-2859/unchecked-transfer-return-value-in-action-dispatcher-and-handlers

Similar to https://github.com/cardstack/card-pay-protocol/pull/177

### DIFF
--- a/contracts/libraries/SafeERC677.sol
+++ b/contracts/libraries/SafeERC677.sol
@@ -8,10 +8,8 @@ library SafeERC677 {
     address to,
     uint256 value,
     bytes memory data
-  ) internal returns (bool) {
+  ) internal {
     bool result = token.transferAndCall(to, value, data);
     require(result, "safeTransferAndCall failed");
-
-    return true;
   }
 }
--- a/contracts/libraries/SafeERC677.sol
+++ b/contracts/libraries/SafeERC677.sol
@@ -1,0 +1,17 @@
+pragma solidity ^0.8.9;
+pragma abicoder v1;
+import "../token/IERC677.sol";
+
+library SafeERC677 {
+  function safeTransferAndCall(
+    IERC677 token,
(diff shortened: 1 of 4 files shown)


contracts/libraries/SafeERC677.sol
```diff
@@ -0,0 +1,17 @@
...
+pragma abicoder v1;
+import "../token/IERC677.sol";
+
+library SafeERC677 {
+  function safeTransferAndCall(
+    IERC677 token,
+    address to,
+    uint256 value,
+    bytes memory data
+  ) internal returns (bool) {
```
alex-cardstack, line 11: I think we should not return anything here to be consistent with `safeTransfer` - it can only ever be true and wastes some small amount of gas to return a value for no reason,.

alex-cardstack (changes requested):
Looks great apart from the unused return bool, thanks!

contracts/libraries/SafeERC677.sol
jurgenwerk: Yeah makes sense. However I had to add the return [here](https://github.com/cardstack/card-pay-protocol/pull/178/commits/ba85994b3f8167c1173d0ab88c3611edbf36ce52#diff-5e88e4612774f782757ca331921061bef2315fbe635859ef415526c150ddc6b8R111) otherwise a couple of PrepaidCard tests break with ` Error: VM Exception while processing transaction: reverted with reason string 'safe transaction was reverted'`.

alex-cardstack: Perfect, we actually have another audit recommendation in general that we stop using pointless return statements like that but fine for this PR!

alex-cardstack (approved):

PR #185, opened 2022-03-28
Rate drift percentage should be 0-1%

jurgenwerk:
Ticket: CS-3465


alex-cardstack (approved):

alex-cardstack:
@habdelra double checking 1% is the limit we want here but code looks good, thanks!

habdelra:
yeah, it's set to 1% in gnosis chain right now, so i think its fine

PR #186, opened 2022-03-29
Add a sane limit to merchantRegistrationFeeInSPEND

jurgenwerk:
Tickets: CS-3470, CS-3468

I suppose:
1.  1000 SPEND ($10) is a sane upper limit for the registration fee?
2. 10000000 (10%) is a sane upper limit for the merchant fee percentage?

### DIFF
--- a/test/RevenuePool-test.js
+++ b/test/RevenuePool-test.js
@@ -257,6 +257,24 @@ contract("RevenuePool", (accounts) => {
           "merchantRegistrationFeeInSPEND is above the maximum"
         );
     });
+
+    it("has a sane upper bound for merchantFeePercentage", async () => {
+      await revenuePool
+        .setup(
+          exchange.address,
+          merchantManager.address,
+          actionDispatcher.address,
+          prepaidCardManager.address,
+          merchantFeeReceiver,
+          10000001,
+          1000,
+          versionManager.address
+        )
+        .should.be.rejectedWith(
+          Error,
+          "merchantFeePercentage is above the maximum"
+        );
+    });
   });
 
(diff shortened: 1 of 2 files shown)


alex-cardstack (approved):
Code looks great, thank you!

test/RevenuePool-test.js
```diff
@@ -239,6 +239,42 @@ contract("RevenuePool", (accounts) => {
...
+
+    it("has a sane upper bound for merchantRegistrationFeeInSPEND", async () => {
+      await revenuePool
+        .setup(
+          exchange.address,
+          merchantManager.address,
+          actionDispatcher.address,
+          prepaidCardManager.address,
+          merchantFeeReceiver,
+          10000000,
```
alex-cardstack, line 251: Nitpick: when passing numbers into solidity calls, it's best practice to pass as a string, i.e. `10000000` (or use [BigNumber](https://docs.ethers.io/v5/api/utils/bignumber/)). This is because javascript's Number type is only precise up to 2^53 and uint256 goes up to 2^256 - for these hardcoded low values not an issue though

jurgenwerk: Noted, thanks!

alex-cardstack:
@habdelra @tintinthong another one where we don't know the limits we should apply

habdelra:
so this is one of those situations we charge a fee in lieu of gas. For networks that have low gas fees yeah, that seems fine, but for a network like mainnet that will be too low. I wonder if a better way to measure this is as a factor of native coin in the network converted to USD--like 1000x native coin.

PR #187, opened 2022-03-29
Document where a GnosisSafe ownership is transferred

jurgenwerk:
Ticket: CS-3358

Would this small change suffice to document places where the gnosis safe ownership is transferred?


alex-cardstack:
@jurgenwerk yes this is good, thank you!

alex-cardstack (approved):

PR #189, opened 2022-04-01
Improve onTokenTransfer hooks documentation

jurgenwerk:
Ticket: CS-3350

This PR aims improve the documentation on how we use the ERC677 `onTokenTransfer` callback in our contracts. It explains the params and offers a short description about its intent and params, and points to the spot in the readme where one can get more details on the handlers. 

~~TODO: checking with @tintinthong to see how to explain the ignored reward related empty param and if we can remove it~~

### DIFF
--- a/contracts/action-handlers/AddRewardRuleHandler.sol
+++ b/contracts/action-handlers/AddRewardRuleHandler.sol
@@ -41,9 +41,26 @@ contract AddRewardRuleHandler is Ownable, Versionable {
     return true;
   }
 
+  /**
+   * @dev onTokenTransfer(ERC677) - this is the ERC677 token transfer callback.
+   *
+   * This adds the reward rule to the reward manager.
+   *
+   * See AddRewardRuleHandler in README for more information.
+   *
+   * @param from the token sender (should be the action dispatcher)
+   * @param data encoded as: (
+   *  address prepaidCard,
+   *  uint256 ???,
+   *  bytes actionData, encoded as: (
+   *    address rewardProgramID,
+   *    bytes blob (hex encoding of rule blob)
+   *    )
+   *  )
+   */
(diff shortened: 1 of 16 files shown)


contracts/action-handlers/AddRewardRuleHandler.sol
```diff
@@ -41,9 +41,26 @@ contract AddRewardRuleHandler is Ownable, Versionable {
...
+   * @dev onTokenTransfer(ERC677) - this is the ERC677 token transfer callback.
+   *
+   * This adds the reward rule to the reward manager.
+   *
+   * See AddRewardRuleHandler in README for more information.
+   *
+   * @param from the token sender (should be the action dispatcher)
+   * @param data encoded as: (
+   *  address prepaidCard,
+   *  uint256 ???,
```
jurgenwerk, line 54: @tintinthong here in [this line](https://github.com/cardstack/card-pay-protocol/blob/main/test/RewardManager-test.js#L486) it says it doesn't matter what this is - how come we can't just remove it?

alex-cardstack (approved):
Excellent work, thank you @jurgenwerk

contracts/action-handlers/AddRewardRuleHandler.sol
tintinthong: Owh this was in the test right. Yea just remove all instance of "it doesn't matter what it is". I think it was me simulating a transferAndCall but the handler doesn't care about the amount.

tintinthong (approved):
LGTM

jurgenwerk:
@alex-cardstack @tintinthong added a commit with the explanation of the ignored param. I think this is now good to merge!

alex-cardstack (approved):

PR #190, opened 2022-04-01
Require 0 token amount where tokens are not needed

jurgenwerk:
Prevents somebody from accidentally transferring tokens to contracts where we don't require any tokens.


alex-cardstack (approved):
This is definitely a good idea to enforce

PR #191, opened 2022-04-07
Add PrepaidCardMarketV2 contract 

jurgenwerk:
Ticket: CS-3511

- [x] Issuer can send tokens to the contract
- [x] Issuer can withdraw tokens from the contract
- [x] Issuer can add SKUs
- [x] Trusted provisioner (will be relay server) can call a method to create a card and transfer it to an EOA by specifying a SKU and target EOA
- [x] Only this contract can call PrepaidManager's `createCardWithIssuer` method
- [x] Fix the V1 market contract tests after changes in the `PrepaidCardManager`
- [x] Add a common interface for v1 and v2 contracts
- [x] Adjust README (Add V2 contract description, change the PrepaidCardManager description)


jurgenwerk:
I'm experiencing an issue when running the test - `Error: Transaction reverted: function call to a non-contract account`

It happens on this line of code in the test: `let gasEstimate = await transferAndCall.estimateGas();`

@habdelra @alexspeller do you maybe know what's going on here? I'm puzzled, `prepaidCardMarketV2` is a contract, why would an error happen where it says that this is not a contract?

jurgenwerk:
@habdelra I think I'm already doing that in this line https://github.com/cardstack/card-pay-protocol/pull/191/files#diff-6b6d90a69f055624761115b00db7b3f1e6ddae545f5be10854567a06ee3dcd23R124

jurgenwerk:
The onTokenTransfer test now passes - I'll start working on token withdraw functionality.

contracts/PrepaidCardManager.sol
```diff
@@ -550,8 +573,39 @@ contract PrepaidCardManager is Ownable, Versionable, Safe {
...
+    address customer,
+    address issuer,
+    address depot, // the issuers safe
+    address from,
+    address token,
+    uint256 issuingTokenAmount,
+    uint256 spendAmount,
+    string memory customizationDID
+  ) internal returns (address) {
+    // TODO: only v2 contract can call this function
```
jurgenwerk, line 586: Should we set an `AddressSet` of trusted addresses that can call this function? And check whether it contains the PrepaidCardMarketV2's address? Or directly check if `from == prepaid card market v2 address`?

contracts/PrepaidCardMarket_v2.sol
```diff
@@ -0,0 +1,304 @@
...
+
+import "hardhat/console.sol";
+
+contract PrepaidCardMarketV2 is Ownable, Versionable, Safe {
+  using EnumerableSetUpgradeable for EnumerableSetUpgradeable.AddressSet;
+  using SafeERC20Upgradeable for IERC677;
+  using SafeERC677 for IERC677;
+
+  struct SKU {
+    address issuerSafe;
```
jurgenwerk, line 26: `issuerSafe` was added so that we can get the safe address in the `getQuantity(bytes32 sku)` method.

I suppose an alternative would be to remove `issuer` and keep `issuerSafe` in the SKU struct, and get the issuer from the issuer mapping in other places in code where we need the issuer. What do you think?

jurgenwerk:
@habdelra @alex-cardstack this is now the first draft of the almost finished contract, I'll be working on adding more unhappy path tests, also there is a TODO missing (https://github.com/cardstack/card-pay-protocol/pull/191#discussion_r857509128). Can you please check if this contract makes sense conceptually, especially the flow of tokens, issuers and owners of prepaid cards? Would also appreciate if you can think of anything that is missing here.

contracts/PrepaidCardManager.sol
```diff
@@ -212,20 +220,34 @@ contract PrepaidCardManager is Ownable, Versionable, Safe {
...
+        from,
+        _msgSender(),
+        amount,
+        issuingTokenAmounts,
+        spendAmounts,
+        customizationDID,
+        marketAddress
+      );
+    } else {
+      createCardWithIssuer(
```
habdelra, line 239: This is a trusted operation--since we need to trust that the caller is forthright in terms of who the issuer is. We should add a require that only the PrepaidCardMarketV2 contract is able to call this (the `from` address). Which also means that the `setup` function needs to be provided with an array of trusted callers for this type of prepaidcard creation (it should be an array to be compatible with future protocol contracts that may want to do this).

contracts/PrepaidCardMarket_v2.sol
habdelra: i think an issuer can have muliple safes that they use to fund inventory but still retain the same SKU--meaning that the issuerSafe should not be part of the SKU. If it's just a lookup that we need, then let's just create a mapping like `issuers`, only it will be in the opposite direction. Since the issuer's safe is not part of the SKU then i would prefer not to store it in this struct.

contracts/PrepaidCardMarket_v2.sol
```diff
@@ -0,0 +1,304 @@
...
+      if (_owners[i] == _issuer) {
+        _foundOwner = true;
+        break;
+      }
+    }
+
+    require(_foundOwner, "issuer is not one of the safe owners");
+
+    balance[from][msg.sender] = balance[from][msg.sender] + amount;
+    issuers[from] = _issuer;
```
habdelra, line 295: we could also add an `issuersSafe` that is the inverse of `issuers` here too in order to help with the `getQuantity` calculations

contracts/PrepaidCardMarket_v2.sol
```diff
@@ -0,0 +1,304 @@
...
+    uint256 faceValue;
+    string customizationDID;
+  }
+
+  address public prepaidCardManagerAddress;
+  address public tokenManager;
+  address public provisioner;
+  address public versionManager;
+  address public exchangeAddress;
+  mapping(address => mapping(address => uint256)) public balance; // issuer safe address -> token -> balance
```
habdelra, line 38: in order to support an issuer using multiple safes to fund their balance, maybe this should be //issuer EOA address -> token -> balance

habdelra: sorry, I take back, i think we should still track this at the safe level, otherwise you might be able to embezzle funds from a safe that has a threshold higher than 1 that you happen to be a part owner of

contracts/PrepaidCardMarket_v2.sol
```diff
@@ -0,0 +1,304 @@
...
+  function provisionPrepaidCard(address customer, bytes32 sku)
+    external
+    returns (bool)
+  {
+    require(!paused, "Contract is paused");
+    require(
+      trustedProvisioners.contains(msg.sender),
+      "Only trusted provisioners allowed"
+    );
+    require(asks[sku] > 0, "Can't provision SKU with 0 askPrice");
```
habdelra, line 138: I think we are missing a require to ensure that there is enough balance for the issuer in order to create the prepaid card

contracts/PrepaidCardMarket_v2.sol
```diff
@@ -0,0 +1,304 @@
...
+      abi.encode(
+        customer,
+        issuingTokenAmounts,
+        spendAmounts,
+        customizationDID,
+        address(0), // marketAddress - we probably don't need it in this case?
+        issuer,
+        issuerSafe
+      )
+    );
```
habdelra, line 177: we should require() that this command is successful. the `safeTransferAndCall()` returns a boolen success if the transfer and call was able to happen. We whould wrap this entire call in `require()`

contracts/PrepaidCardMarket_v2.sol
```diff
@@ -0,0 +1,304 @@
...
+
+    balance[issuerSafe][token] -= priceToCreatePrepaidCard;
```
habdelra, line 179: we should deduct this _before_ the safeTransferAndCall in order to prevent reentrancy

contracts/PrepaidCardMarket_v2.sol
```diff
@@ -0,0 +1,304 @@
...
+
+    return true;
```
habdelra, line 181: I think we are missing a ProvisionPrepaidCard event, can you make sure that is part of the unit tests too?

contracts/PrepaidCardMarket_v2.sol
```diff
@@ -0,0 +1,304 @@
...
+
+    IERC677(token).safeTransferAndCall(
+      prepaidCardManagerAddress,
+      priceToCreatePrepaidCard,
+      abi.encode(
+        customer,
+        issuingTokenAmounts,
+        spendAmounts,
+        customizationDID,
+        address(0), // marketAddress - we probably don't need it in this case?
```
habdelra, line 173: confirm--this should always be `address(0)` in this case

contracts/PrepaidCardMarket_v2.sol
habdelra: also, it would be interesting if this issuer had multiple safes with balances, and that if one of their balances for the safes became depleted that we could use their other safe balances. i don't think that is an immediate requirement, but that is a nice to have.

contracts/PrepaidCardMarket_v2.sol
habdelra: as I look deeper into the logic i see a conundrum in that if an issuer has multiple safes, you won't know which one to pull from (ideally any might be valid--but figuring out which and being able to roll over into different safe balances could be tricky). Given that, I'm ok with having the issuer safe as part of the SKU struct, but probably we should not make it part of the SKU hash itself, so that eventually we could support the ability to have multiple safes fund all the prepaid cards for an issuer.

contracts/PrepaidCardMarket_v2.sol
```diff
@@ -0,0 +1,304 @@
...
+    return balance[issuerSafe][token] / price;
+  }
+
+  function setAsk(
+    address issuer,
+    bytes32 sku,
+    uint256 askPrice
+  ) external returns (bool) {
+    require(skus[sku].issuer != address(0), "Non-existent SKU");
+    require(skus[sku].issuer == issuer, "SKU not owned by issuer");
```
habdelra, line 204: we also need a require (and a test) that prevents a random person from setting the ask. these will be safe txns, so the msg.sender should be the issuer's safe for the SKU that you are trying to set.

contracts/PrepaidCardMarket_v2.sol
```diff
@@ -0,0 +1,304 @@
...
+  function setPaused(bool _paused) external onlyOwner {
+    paused = _paused;
+    emit PausedToggled(_paused);
+  }
+
+  function getTrustedProvisioners() external view returns (address[] memory) {
+    return trustedProvisioners.values();
+  }
+
+  function _addTrustedProvisioner(address _token) internal {
```
habdelra, line 119: let's add an `onlyOwner` to this to be defensive--it's very good for documentation as well.

contracts/PrepaidCardMarket_v2.sol
habdelra: or actually, I think the point of _safe_TransferAndCall, the thing that makes it "safe", is that it wraps this for you. so never mind.

contracts/PrepaidCardMarket_v2.sol
```diff
@@ -0,0 +1,304 @@
...
+
+    IERC677(token).safeTransfer(msg.sender, amount);
+
+    emit TokensWithdrawn(msg.sender, _issuer, token, amount);
+  }
+
+  function onTokenTransfer(
+    address payable from, // safe address
+    uint256 amount,
+    bytes calldata data
```
habdelra, line 267: let's document what the `calldata` is (encoded address specifying the issuer) to make it easy for folks reading the code

test/PrepaidCardMarket_v2-test.js
```diff
@@ -0,0 +1,504 @@
...
+
+      expect(
+        (await daicpxdToken.balanceOf(createPrepaidCardEvent.card)).toString()
+      ).to.equal(toTokenUnit(50).toString());
+
+      expect(
+        await prepaidCardManager.getPrepaidCardOwner(
+          createPrepaidCardEvent.card
+        )
+      ).to.equal(customer);
```
habdelra, line 486: let's also add an assertion that an event was fired

contracts/PrepaidCardMarket_v2.sol
jurgenwerk: I agree, added this in https://github.com/cardstack/card-pay-protocol/pull/191/commits/bdb3e0d6f6766f8f9c1a914032cf0a57b3b8a808

contracts/PrepaidCardMarket_v2.sol
```diff
@@ -0,0 +1,304 @@
...
+    string memory customizationDID = skus[sku].customizationDID;
+    uint256 priceToCreatePrepaidCard = prepaidCardManager.priceForFaceValue(
+      token,
+      faceValue
+    ); // in Wei
+
+    uint256[] memory issuingTokenAmounts = new uint256[](1);
+    uint256[] memory spendAmounts = new uint256[](1);
+
+    issuingTokenAmounts[0] = Exchange(exchangeAddress).convertFromSpend(
```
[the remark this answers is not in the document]
jurgenwerk: Thanks, fixed in https://github.com/cardstack/card-pay-protocol/pull/191/commits/72389f7b36771e8d0f678ee92512b89f60bb559e

contracts/PrepaidCardMarket_v2.sol
jurgenwerk: 👍

contracts/PrepaidCardMarket_v2.sol
jurgenwerk: Reentrancy addressed in https://github.com/cardstack/card-pay-protocol/pull/191/commits/c6d7a3883265817dc6798f15e4ee01047a57448a

contracts/PrepaidCardMarket_v2.sol
```diff
@@ -0,0 +1,304 @@
...
+  function addSKU(
+    uint256 faceValue,
+    string memory customizationDID,
+    address token
+  ) external returns (bool) {
+    require(faceValue > 0, "Face value must be greater than 0");
+    require(token != address(0), "Token address must be set");
+
+    address _issuer = issuers[msg.sender];
+    require(_issuer != address(0), "Issuer not found.");
```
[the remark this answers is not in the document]
jurgenwerk: Agreed. Fixed in https://github.com/cardstack/card-pay-protocol/pull/191/commits/102d5f514d8966c00f42ea851879f939ecbb5463

contracts/PrepaidCardMarket_v2.sol
jurgenwerk: Added in https://github.com/cardstack/card-pay-protocol/pull/191/commits/7f6f29ca8720176bf2569a4326b8e89d9b961f6f

test/PrepaidCardMarket_v2-test.js
```diff
@@ -0,0 +1,504 @@
...
+    });
+
+    it(`rejects when contract is paused`, async function () {
+      await prepaidCardMarketV2.setPaused(true);
```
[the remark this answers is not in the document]
jurgenwerk: Fixed in https://github.com/cardstack/card-pay-protocol/pull/191/commits/1624824e2d57e201d5f12052e98f16e11729c085

test/PrepaidCardMarket_v2-test.js
```diff
@@ -0,0 +1,504 @@
...
+      ).to.equal(toTokenUnit(50).toString());
+
+      expect(
+        await prepaidCardManager.getPrepaidCardOwner(
+          createPrepaidCardEvent.card
+        )
+      ).to.equal(customer);
+    });
+
+    it(`rejects when contract is paused`, async function () {
```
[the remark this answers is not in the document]
jurgenwerk: Agree. Added in https://github.com/cardstack/card-pay-protocol/pull/191/commits/1624824e2d57e201d5f12052e98f16e11729c085

contracts/PrepaidCardMarket_v2.sol
jurgenwerk: Do you have any suggestions on what would be the most proper way of emitting this in the `PrepaidCardMarketV2`? It doesn't look so straightforward to me because when we send the tokens using `IERC677(token).safeTransferAndCall` in `provisionPrepaidCard` method, we don't get the prepaid card address back, meaning we can not emit `ProvisionedPrepaidCard(prepaidCard, customer, sku, askPrice)` at this spot.

So I'm wondering where would be good to emit this event. One option is to do it in the `PrepaidCardManager` in the `onTokenTransfer` hook after creating the card, but for that we'd also need to send the `sku` and `askPrice` in the `data`. One other option I see is to provide a `onPrepaidCardCreate` callback function in the V2 contract which the manager would call. Is it maybe possible to somehow query for the newly created prepaid card address in the end of the `provisionPrepaidCard` function in the V2 contract?

habdelra: That’s fine to not have the prepaid card address. The prepaid card manager will emit an event with this address in the same txn that the subgraph can correlate . This can be an event that has different args. Specifically you can include the SKU and the prepaid card owner. It is important though that any publicly facing function always emits an event—it absolutely something that we are audited for.

contracts/PrepaidCardMarket_v2.sol
```diff
@@ -0,0 +1,322 @@
...
+
+    emit AskSet(issuer, skus[sku].issuingToken, sku, askPrice);
+    return true;
+  }
+
+  function addSKU(
+    uint256 faceValue,
+    string memory customizationDID,
+    address token
+  ) external returns (bool) {
```
habdelra, line 221: this is also only called by prepaid cards not generic safes You'll need an `onlyHandlers` modifer here like we do in the V1 version of this https://github.com/cardstack/card-pay-protocol/blob/main/contracts/PrepaidCardMarket.sol#L192

habdelra:
we also might wanna make an interface that both the PrepaidCardMarket V1 and V2 implement so that from teh action handlers we can be sure that the functions we are trying to call exist on the market contracts

jurgenwerk:
@habdelra regarding changing the `addSKU` and `setAsk` so that only handlers can call them - I see there's a `SetPrepaidCardAskHandler` (looks like we can use this using `setPrepaidCardAsk` in test helpers) but there is no handler to add a SKU. Should we create one?

habdelra:
Oh, yeah good point in V1 the sku is derived automatically based on placing cards into the inventory. Yeah you’ll need to make a new action handle for setting the sku.  It can follow the same approach as the action handle for setting the ask price.

contracts/PrepaidCardMarket_v2.sol
jurgenwerk: Now that we added `onlyHandlers` here - does this suffice to satisfy this requirement? The handler will already check `require(issuer == owner, "only issuer can set market inventory");`

habdelra: Yes, I agree.

jurgenwerk:
@habdelra @alex-cardstack This is now ready for another review. I believe I addressed all the concerns, can you take another look?

@habdelra if I remember correctly we were discussing making a common interface for both market contracts, but there are some methods in the current interface (`IPrepaidCardMarket`) which the V2 doesn't have (i.e. `setItem`, `removeItems`). I suppose we can implement empty stubs in the V2, or remove them from the interface. Do you see any benefit in having both contracts conformed to the same interface? 

Also, the number of commits in the PR got quite large. I'll be doing some squashing next, also still need to document the new handler and the V2 contract in the readme.

contracts/PrepaidCardMarketV2.sol
```diff
@@ -0,0 +1,361 @@
...
+    string customizationDID,
+    bytes32 sku
+  );
+  event AskSet(
+    address issuer,
+    address issuingToken,
+    bytes32 sku,
+    uint256 askPrice
+  );
+  event PrepaidCardProvisioned(address owner, bytes32 sku);
```
backspace, line 76: Will `sku` be something we can use to get the address of the newly-provisioned prepaid card? I’m looking at reacting to the `PrepaidCardProvisioned` event, we’d like to be able to have the card’s address.

habdelra: no it's an entirely different thing. the address of the prepaid card is part of the event that the PrepaidCardManager emits in this same txn. This will be queryable from the subgraph based on the reservation table's provisioning txn hash. However we still need to do the subgraph work to capture the new events from the V2 market contract before we can tell you what the query is.

contracts/PrepaidCardManager.sol
```diff
@@ -142,6 +144,7 @@ contract PrepaidCardManager is Ownable, Versionable, Safe {
     uint256 _minAmount,
     uint256 _maxAmount,
     address[] calldata _contractSigners,
+    address[] calldata _trustedCallersForCreatingPrepaidCardsWithIssuer,
```
habdelra, line 147: make sure to update the configuration scripts to set this: https://github.com/cardstack/card-pay-protocol/blob/main/scripts/deploy/config/default/PrepaidCardManager.ts#L22-L59

habdelra (commented):
You need to also provide configuration scripts for the new V2 Market contract. Follow the examples here https://github.com/cardstack/card-pay-protocol/tree/main/scripts/deploy/config/default

contracts/PrepaidCardManager.sol
```diff
@@ -624,7 +677,7 @@ contract PrepaidCardManager is Ownable, Versionable, Safe {
     );
 
     if (marketAddress != address(0)) {
-      IPrepaidCardMarket(marketAddress).setItem(owner, card);
+      PrepaidCardMarket(marketAddress).setItem(owner, card);
```
habdelra, line 680: good call :+1: this is not compatible between the 2 market contracts

contracts/action-handlers/AddPrepaidCardSKUHandler.sol
```diff
@@ -0,0 +1,95 @@
...
+  address public versionManager;
+
+  event Setup();
+
+  function setup(
+    address _actionDispatcher,
+    address _prepaidCardManager,
+    address _tokenManagerAddress,
+    address _versionManager
+  ) external onlyOwner returns (bool) {
```
habdelra, line 26: Need to add a configuration for this contract in https://github.com/cardstack/card-pay-protocol/tree/main/scripts/deploy/config/default

habdelra: Additionally you'll need to configure ActionDispatcher for this new action too https://github.com/cardstack/card-pay-protocol/blob/main/scripts/deploy/config/default/ActionDispatcher.ts

contracts/action-handlers/AddPrepaidCardSKUHandler.sol
```diff
@@ -0,0 +1,95 @@
...
+
+    (, , bytes memory actionData) = abi.decode(data, (address, uint256, bytes));
+
+    (
+      uint256 faceValue,
+      string memory customizationDID,
+      address tokenAddress,
+      address marketAddress,
+      address issuerSafe
+    ) = abi.decode(actionData, (uint256, string, address, address, address));
```
habdelra, line 79: probably we should also require that the owner of the prepaid card is the same as the owner of the issuer safe

contracts/PrepaidCardMarketV2.sol
```diff
@@ -0,0 +1,359 @@
...
+      prepaidCardManagerAddress
+    );
+
+    address token = skus[sku].issuingToken;
+    uint256 faceValue = skus[sku].faceValue;
+    address issuerSafe = skus[sku].issuerSafe;
+
+    uint256 price = prepaidCardManager.priceForFaceValue(token, faceValue);
+
+    return balance[issuerSafe][token] / price;
```
alex-cardstack, line 220:
```suggestion
    // Division in solidity rounds towards zero, so this calculation won't overestimate the quantity available
    // https://docs.soliditylang.org/en/latest/types.html#division
    return balance[issuerSafe][token] / price;
```

I verified this then figured I'd suggest it as a comment

contracts/PrepaidCardMarketV2.sol
```diff
@@ -0,0 +1,359 @@
...
+    address issuer,
+    address token,
+    uint256 faceValue,
+    string memory customizationDID
+  ) public pure override returns (bytes32) {
+    return
+      keccak256(abi.encodePacked(issuer, token, faceValue, customizationDID));
+  }
+
+  function withdrawTokens(uint256 amount, address token) external {
```
alex-cardstack, line 276:
```suggestion
  // Note: msg.sender is issuer safe
  function withdrawTokens(uint256 amount, address token) external {
```

contracts/PrepaidCardMarketV2.sol
```diff
@@ -0,0 +1,359 @@
...
+   * @param data encoded as (
+   *  address issuer (issuer's address)
+   * )
+   */
+  function onTokenTransfer(
+    address payable from, // safe address
+    uint256 amount,
+    bytes calldata data
+  ) external returns (bool) {
+    // Only CARD.CPXD, DAI.CPXD are accepted
```
alex-cardstack, line 324:
```suggestion
    // Only whitelisted tokens are accepted
```

Not important, but FYI, the .CPXD postfix will change on other networks, and other tokens might be added too

contracts/PrepaidCardMarketV2.sol
alex-cardstack, line 276: I have had a thought that I'm not sure of the answer to:

What are the implications if an issue safe changes owners.

* Is it still possible to withdraw tokens?
* are there other issues for the other functionality of the contract?

Not sure of the answer to either question

alex-cardstack (changes requested):
Looking really good, a few minor bits left

contracts/PrepaidCardManager.sol
```diff
@@ -215,44 +227,81 @@ contract PrepaidCardManager is Ownable, Versionable, Safe {
...
-      address marketAddress
-    ) = abi.decode(data, (address, uint256[], uint256[], string, address));
+      address marketAddress,
+      address issuer,
+      address issuerSafe
+    ) = abi.decode(
+        data,
+        (address, uint256[], uint256[], string, address, address, address)
+      );
+
```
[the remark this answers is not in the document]
habdelra: So i noticed something interesting when I was looking over the gnosis safe owner stuff recently--they actually use an `address(0x1)` as a special constant https://github.com/safe-global/safe-contracts/blob/main/contracts/base/OwnerManager.sol#L13. Maybe we could do something like that too? perhaps we can treat `address(0x1)` as type of "null" address to work around these issues? (and then update the SDK accordingly?)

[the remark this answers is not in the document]
habdelra: additionally, the SDK also has support for preventing use from using an older version of the SDK against our newer contracts. The SDK is configured with a set of protocol versions that it is compatible with. the SDK will work for any future patch version of a protocol version that it is compatible with. However, the SDK will not work with a future minor version of the protocol version that it is compatible with. So long as this is released as a minor version of teh SDK, the existing SDK's will not be able to talk to this contract using their current logic. Only SDK's that can been configured to talk to this specific minor version will be able to call this contract--meaning that the calldata should be assembled with all the params that need to be specified in the next version of the SDK that will support this minor version.

contracts/PrepaidCardManager.sol
```diff
@@ -215,44 +227,81 @@ contract PrepaidCardManager is Ownable, Versionable, Safe {
...
+        from, // depot
+        _msgSender(), // token
+        amount,
+        issuingTokenAmounts,
+        spendAmounts,
+        customizationDID,
+        marketAddress
+      );
+    } else {
+      // In case when issuer and issuerSafe are provided, it means the tokens are being
```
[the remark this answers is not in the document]
habdelra: i think we're ok so long as we release this as a new minor version--the older SDK will not be permitted to talk to this contract if it's a new minor version.

habdelra:
Also, i think the smoke test failures are because we haven't update the config scripts that i pointed out in my review

contracts/action-handlers/AddPrepaidCardSKUHandler.sol
jurgenwerk: Nice catch - addressed in https://github.com/cardstack/card-pay-protocol/pull/191/commits/0b6b8179edb359d5098c4ea8cb875c8ab7584e5e

test/utils/constant/eventABIs.js
```diff
@@ -284,6 +284,126 @@ const eventABIs = {
...
+      "SkuAdded(address,address,uint256,string,bytes32)"
+    ),
+    abis: [
+      {
+        type: "address",
+        name: "issuer",
+      },
+      {
+        type: "address",
+        name: "token",
```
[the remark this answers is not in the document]
jurgenwerk: Nice catch, fixed.

contracts/PrepaidCardManager.sol
```diff
@@ -170,6 +173,15 @@ contract PrepaidCardManager is Ownable, Versionable, Safe {
 
       contractSigners.add(_contractSigners[i]);
     }
+    for (
+      uint256 i = 0;
+      i < _trustedCallersForCreatingPrepaidCardsWithIssuer.length;
```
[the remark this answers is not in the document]
jurgenwerk: Added in https://github.com/cardstack/card-pay-protocol/pull/191/commits/6e6185da8c9f63dc22887b5f2d11993dd502ed2f

scripts/deploy/config/default/PrepaidCardMarketV2.ts
```diff
@@ -0,0 +1,42 @@
...
+        name: "tokenManager",
+        value: address("TokenManager"),
+      },
+      {
+        name: "actionDispatcher",
+        value: address("ActionDispatcher"),
+      },
+      {
+        name: "getTrustedProvisioners",
+        value: [PREPAID_CARD_PROVISIONER],
```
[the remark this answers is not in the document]
habdelra: or my bad--i thought this was the _PrepaidCardManager_ never mind

contracts/PrepaidCardMarketV2.sol
```diff
@@ -0,0 +1,375 @@
...
+    address issuerSafe;
+    address issuer;
+    address issuingToken;
+    uint256 faceValue;
+    string customizationDID;
+  }
+
+  address public prepaidCardManagerAddress;
+  address public tokenManager;
+  address public provisioner;
```
habdelra, line 38: why do we have both this and `trustedProvisioners`? aren't they both the same thing?

habdelra: (also i think a provisioner is assumed to be trusted--there is no such thing as an "untrusted provisioner", probably just calling this `provisioners` is fine).

jurgenwerk: Actually I don't see a good reason to have both, unless we'll have many provisioners? If not we can remove the array and set the relayer as provisioner in tests

jurgenwerk: Actually I don't see a good reason to have both, I think we can remove the `provisioner` and rename `trustedProvisioners` to `provisioners`

alex-cardstack (approved):
Looking good!

contracts/PrepaidCardMarketV2.sol
jurgenwerk: Done!

habdelra (approved):

jurgenwerk:
@habdelra @alex-cardstack I added some more documentation to the README, can you check if it makes sense and if you think anything is missing?

alex-cardstack (approved):
Documentation changes look accurate to me, with one question about a paragraph that is not important

README.md
```diff
@@ -54,8 +68,11 @@ The `PrepaidCardManager` contract stores signers that are authorized to sign saf
...
 
-A future update to this contract will provide the ability for EOAs to directly purchase Prepaid Card safes from the PrepaidCardMarket using native coin and/or CPXD tokens. In that future scenario we would leverage the "ask price" that can be set against a SKU to be able to purchase Prepaid Card safes directly. To support this eventual feature we have added the ability to set an "ask price" for SKUs ni the inventory, however, capability to directly purchase a Prepaid Card from this contract is still TBD.
+A future update to this contract will provide the ability for EOAs to directly purchase Prepaid Card safes from the PrepaidCardMarket using native coin and/or CPXD tokens. In that future scenario we would leverage the "ask price" that can be set against a SKU to be able to purchase Prepaid Card safes directly. To support this eventual feature we have added the ability to set an "ask price" for SKUs in the inventory, however, capability to directly purchase a Prepaid Card from this contract is still TBD.
```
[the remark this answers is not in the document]
habdelra: this is correct. there is currently a desire that the market contracts be able to also accept on-chain payments with supported tokens (not necessarily CPXD tokens) in exchange for prepaid cards (or in the future some managed safe)

PR #193, opened 2022-04-12
Fix contract verification using nomic provided verification plugin

alex-cardstack:
So, after trying to reproduce verification on etherscan using output we were trying to verify on blockscout, it turns out the problem was not on blockscouts end after all.

The root cause was the sol-merger package, which somehow manages to produce incorrectly merged solidity files, so we were trying to verify the wrong code.

Fortunately for us, nomiclabs have started supporting blockscout in the `@nomiclabs/hardhat-etherscan` package, so we can just use that, it seems to use standard solc input json so avoids the need for flattening altogether.

Lots of code deleted and all our contracts are now verified on xdai!

### DIFF
--- a/hardhat.config.ts
+++ b/hardhat.config.ts
@@ -9,6 +9,7 @@ import "@openzeppelin/hardhat-upgrades";
 import "hardhat-contract-sizer";
 import "hardhat-watcher";
 import "solidity-coverage";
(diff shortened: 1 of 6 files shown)


lukemelia (approved):
👏 👏 👏

PR #197, opened 2022-05-03
Add setPrepaidCardUsed to reward actions 

tintinthong:


tintinthong:
@alex-cardstack there is the smoke test failing. I believe because its not upgradeable. How should I resolve this? I dont think the handers carry any important state in this case.

jurgenwerk (approved):
LGTM!

contracts/action-handlers/PayRewardTokensHandler.sol
```diff
@@ -21,22 +22,26 @@ contract PayRewardTokensHandler is Ownable, Versionable {
   address public actionDispatcher;
   address public tokenManagerAddress;
   address public rewardPoolAddress;
+  address public prepaidCardManagerAddress;
```
alex-cardstack, line 25: This is changing the order of the storage variables - the new variable must go at the end

contracts/action-handlers/RegisterRewardProgramHandler.sol
```diff
@@ -21,13 +22,15 @@ contract RegisterRewardProgramHandler is Ownable, Versionable {
   address public exchangeAddress;
   address public tokenManagerAddress;
   address public rewardManagerAddress;
+  address public prepaidCardManagerAddress;
```
alex-cardstack, line 25: Storage order changed - this must go at the end after versionManager

alex-cardstack (changes requested):

tintinthong:
@alex-cardstack can this go in?

PR #198, opened 2022-05-10
Ensure withdraw works if safe owners changed

alex-cardstack:
It was a pain to write the test with the safe transaction to swap owners, but it turns out it works fine without changing the contract


jurgenwerk (approved):
LGTM!

Issue #206, opened 2022-06-02
Fix code scanning alert - High severity No 2


FadhlanR:
Tracking issue for:
- [ ] https://github.com/cardstack/card-pay-protocol/security/code-scanning/2

Issue #205, opened 2022-06-02
Fix code scanning alert - High severity No 1


FadhlanR:
Tracking issue for:
- [ ] https://github.com/cardstack/card-pay-protocol/security/code-scanning/3

PR #207, opened 2022-06-02
Fix high severity issues

FadhlanR:
Ticket: CS-3982
Issues: #205, #206


jurgenwerk (approved):
I think this is good. We did the same thing in other places as well but looks like we forgot to change it here in these 2 places.

PR #230, opened 2022-06-10
Adjust MerchantManager contract so that trusted registrars can also call registerMerchant

FadhlanR:
Tickets: CS-4025

- [x] Add an address set named merchantRegistrars which will contain addresses (relay server in this case).
- [x] Modify registerMerchant so that registrars can perform this function too.
- [x] Adjust initializers so that the address of the relay gets added in the setup function (should be similar to implementation of _addProvisioner in the PrepaidCardMarketV2 market).


contracts/MerchantManager.sol
```diff
@@ -65,9 +74,21 @@ contract MerchantManager is Ownable, Versionable, Safe {
     return merchants[merchant].values();
   }
 
+  function getMerchantRegistrars() external view returns (address[] memory) {
+    return merchantRegistrars.values();
+  }
+
+  function _addMerchantRegistrars(address provisionerAddress)
```
jurgenwerk, line 81: Did you mean `merchantRegistrarAddress`?

test/RevenuePool-test.js
```diff
@@ -681,6 +684,33 @@ contract("RevenuePool", (accounts) => {
       ).to.equal("did:cardstack:56d6fc54-d399-443b-8778-d7e4512d3a49-xx");
     });
 
+    it("allows the merchant registrar to directly add a merchant", async () => {
+      let merchantTx = await merchantManager.registerMerchant(
+        merchantThree,
```
jurgenwerk, line 689: It's confusing to me that the test starts with `merchantThree`. This naming implies that there are merchants 1 and 2 in play here but they are not. However I think I see the problem - the test assertions are all sharing state set in the `before` hook. I think it would be better to change that to `beforeEach` so that every assertion gets its own fresh state, and you could simply use the `merchant` variable, but maybe out of scope for this task. Otherwise I suggest you start the test with: `let merchant = accounts[N]` like a couple of other asssertions do.

jurgenwerk (changes requested):
I think it would be good to add tests for adding and removing registrars, like here: https://github.com/cardstack/card-pay-protocol/blob/main/test/PrepaidCardMarketV2-test.js#L219-L261

habdelra (commented):
Let's also add a negative test where we assert that the contract reverts when a non-merchant registrar tries to create a merchant

FadhlanR:
@habdelra I think for negative test already covered by this existing testcase, should I just renaming it?
```
it("rejects a non-contract owner directly adding a merchant", async () => {
      await merchantManager
        .registerMerchant(
          accounts[9],
          "did:cardstack:56d6fc54-d399-443b-8778-d7e4512d3a49",
          { from: accounts[9] }
        )
        .should.be.rejectedWith(
          Error,
          "caller is not a registered action handler nor an owner"
        );
    });
```

habdelra:
ah yu're right--i missed it because it wasn't a change for this PR. looks good!

habdelra (approved):
Pending the issues @jurgenwerk raised, i approve this

test/RevenuePool-test.js
```diff
@@ -302,7 +303,51 @@ contract("RevenuePool", (accounts) => {
...
+        versionManager.address
+      );
+      expect(await merchantManager.getMerchantRegistrars()).to.have.members([
+        relayer,
+      ]);
+    });
+  });
+
+  describe("remove merchant manager", () => {
+    it("should removed trusted merchant registrars", async () => {
```
jurgenwerk, line 322: This sounds better: 

"removing merchant registrars"
"can remove merchant registrars"

jurgenwerk (approved):
Looking good, thanks for tidying up the commits!

test/RevenuePool-test.js
FadhlanR: Updated

PR #231, opened 2022-06-14
Support disabling merchant safe

FadhlanR:
Ticket: CS-4063

- [x] Add function disableSafe(address _merchantSafe)
- [x] Only owner or merchant registrar (relayer) can call disableSafe
- [x] Use an address set to keep track of which safes are considered disabled
- [x] Add the inverse function (to enable it)
- [x] Add a public getter so we can see which safes are disabled
- [x] In the RevenuePool contract, in the claimRevenue function, require that the safe is not disabled. This is so that we prevent the user to retrieve any of the funds on their prepaid cards

### DIFF
--- a/test/RevenuePool-test.js
+++ b/test/RevenuePool-test.js
@@ -797,6 +797,105 @@ contract("RevenuePool", (accounts) => {
     });
   });
 
+  describe("disable merchant safe", () => {
+    let merchantSafe;
+    before(async () => {
+      await merchantManager.setup(
+        actionDispatcher.address,
+        gnosisSafeMasterCopy.address,
+        proxyFactory.address,
+        [relayer],
+        versionManager.address
+      );
+
+      let merchant = accounts[11];
+      let merchantTx = await merchantManager.registerMerchant(
+        merchant,
+        "did:cardstack:56d6fc54-d399-443b-8778-d7e4512d3a49-xz",
+        { from: relayer }
+      );
+      let merchantCreation = await getParamsFromEvent(
+        merchantTx,
+        eventABIs.MERCHANT_CREATION,
+        merchantManager.address
+      );
+      merchantSafe = merchantCreation[0]["merchantSafe"];
+    });
+
+    afterEach(async () => {
+      await merchantManager.enableSafe(merchantSafe);
+      expect(await merchantManager.getDisabledMerchantSafes()).to.be.empty;
+    });
+
+    it("allows owner to disable merchant safe", async () => {
+      await merchantManager.disableSafe(merchantSafe, { from: owner });
+      expect(await merchantManager.getDisabledMerchantSafes()).to.have.members([
+        merchantSafe,
+      ]);
+    });
+
+    it("allows registrar to disable merchant safe", async () => {
+      await merchantManager.disableSafe(merchantSafe, { from: relayer });
+      expect(await merchantManager.getDisabledMerchantSafes()).to.have.members([
+        merchantSafe,
+      ]);
+    });
+
+    it("does not allow non-owner nor non-registrar to disable merchant safe", async () => {
+      await merchantManager
+        .disableSafe(merchantSafe, { from: accounts[11] })
+        .should.be.rejectedWith(
+          Error,
+          "caller is not an owner nor a registrar"
+        );
+    });
+  });
+
+  describe("enable merchant safe", () => {
+    let merchantSafe;
+    before(async () => {
+      await merchantManager.setup(
+        actionDispatcher.address,
+        gnosisSafeMasterCopy.address,
+        proxyFactory.address,
+        [relayer],
+        versionManager.address
+      );
+
+      let merchant = accounts[13];
+      let merchantTx = await merchantManager.registerMerchant(
+        merchant,
+        "did:cardstack:56d6fc54-d399-443b-8778-d7e4512d3a49-xz",
+        { from: relayer }
+      );
+      let merchantCreation = await getParamsFromEvent(
+        merchantTx,
+        eventABIs.MERCHANT_CREATION,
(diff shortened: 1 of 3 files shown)


test/RevenuePool-test.js
```diff
@@ -797,6 +797,109 @@ contract("RevenuePool", (accounts) => {
...
+      expect(await merchantManager.getDisabledMerchantSafes()).to.have.empty;
+    });
+
+    it("does not allow non-owner nor non-registrar to enable merchant safe", async () => {
+      await merchantManager
+        .enableSafe(merchantSafe, { from: accounts[11] })
+        .should.be.rejectedWith(
+          Error,
+          // the real revert reason is behind the gnosis safe execTransaction
+          // boundary, so we just get this generic error
```
alex-cardstack, line 897: This comment is actually incorrect here, the real revert reason is present I think

alex-cardstack (approved):
Just the incorrect comment I noticed here

test/RevenuePool-test.js
```diff
@@ -1520,6 +1623,31 @@ contract("RevenuePool", (accounts) => {
...
+      );
+      // reverts are trigged via the gas estimation, so we'll never get far
+      // enough to actually issue the execTransaction on the safe
+      await claimRevenue
+        .estimateGas({ from: merchantSafe })
+        .should.be.rejectedWith(Error, "merchant safe is disabled");
+
+      //re-enabled merchant safe
+      await merchantManager.enableSafe(merchantSafe);
+      expect(await merchantManager.getDisabledMerchantSafes()).to.have.empty;
```
jurgenwerk, line 1646: `to.be.empty`

test/RevenuePool-test.js
jurgenwerk: Yeah the comment is wrong, we don't use a gnosis safe transaction for this function.

test/RevenuePool-test.js
```diff
@@ -1520,6 +1623,31 @@ contract("RevenuePool", (accounts) => {
...
+      expect(await merchantManager.getDisabledMerchantSafes()).to.have.members([
+        merchantSafe,
+      ]);
+
+      let amount = toTokenUnit(1);
+      let claimRevenue = revenuePool.contract.methods.claimRevenue(
+        daicpxdToken.address,
+        amount
+      );
+      // reverts are trigged via the gas estimation, so we'll never get far
```
jurgenwerk, line 1638: An alternative is using `.call`, like here: https://github.com/cardstack/card-pay-protocol/blob/main/test/PrepaidCardMarketV2-test.js#L337. Just FYI, no real need to fix this.

test/RevenuePool-test.js
```diff
@@ -797,6 +797,109 @@ contract("RevenuePool", (accounts) => {
...
+    });
+
+    it("allows owner to enable merchant safe", async () => {
+      await merchantManager.enableSafe(merchantSafe, { from: owner });
+      expect(await merchantManager.getDisabledMerchantSafes()).to.have.empty;
+    });
+
+    it("allows registrar to enable merchant safe", async () => {
+      await merchantManager.enableSafe(merchantSafe, { from: relayer });
+      expect(await merchantManager.getDisabledMerchantSafes()).to.have.empty;
```
jurgenwerk, line 888: `to.be.empty`

test/RevenuePool-test.js
```diff
@@ -797,6 +797,109 @@ contract("RevenuePool", (accounts) => {
...
+        merchantTx,
+        eventABIs.MERCHANT_CREATION,
+        merchantManager.address
+      );
+      merchantSafe = merchantCreation[0]["merchantSafe"];
+    });
+
+    it("allows owner to enable merchant safe", async () => {
+      await merchantManager.enableSafe(merchantSafe, { from: owner });
+      expect(await merchantManager.getDisabledMerchantSafes()).to.have.empty;
```
jurgenwerk, line 883: `to.be.empty`

test/RevenuePool-test.js
```diff
@@ -797,6 +797,109 @@ contract("RevenuePool", (accounts) => {
...
+        merchantTx,
+        eventABIs.MERCHANT_CREATION,
+        merchantManager.address
+      );
+      merchantSafe = merchantCreation[0]["merchantSafe"];
+    });
+
+    afterEach(async () => {
+      await merchantManager.enableSafe(merchantSafe);
+      expect(await merchantManager.getDisabledMerchantSafes()).to.have.empty;
```
jurgenwerk, line 827: `to.be.empty`

test/RevenuePool-test.js
FadhlanR: removed the comment

test/RevenuePool-test.js
```diff
@@ -1520,6 +1623,31 @@ contract("RevenuePool", (accounts) => {
...
+      await claimRevenue
+        .estimateGas({ from: merchantSafe })
+        .should.be.rejectedWith(Error, "merchant safe is disabled");
+
+      //re-enabled merchant safe
+      await merchantManager.enableSafe(merchantSafe);
+      expect(await merchantManager.getDisabledMerchantSafes()).to.have.empty;
+      await claimRevenue
+        .estimateGas({ from: merchantSafe })
+        .should.not.rejectedWith();
```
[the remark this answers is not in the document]
FadhlanR: updated

test/RevenuePool-test.js
FadhlanR: updated

test/RevenuePool-test.js
FadhlanR: updated

jurgenwerk (approved):

PR #236, opened 2022-06-30
Recreate revenue pool as safe (Tank)

FadhlanR:
This is the result of a spike process to recreate revenue pools as a safe by leveraging the safe module and guard. Users can directly send tokens to the merchant safe (jar) without worrying about losing pool capabilities. The Tank module helps users to record the locking period of tokens they send to merchant safe (jars). Jar Guard will ensure that only tokens that have been released and exceed the locking period merchants can spend from the jar.

### DIFF
--- a/.solhint.json
+++ b/.solhint.json
@@ -2,7 +2,14 @@
   "extends": "solhint:recommended",
   "rules": {
     "max-states-count": "off",
+    "not-rely-on-time": "off",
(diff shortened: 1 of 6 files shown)


alex-cardstack:
This is looking really good. There would be a few minor comments if we want to merge this but for the purposes of the spike it is great to see and answers the questions well

PR #237, opened 2022-07-08
Fix params sent to CreatePrepaidCard event 

jurgenwerk:
Ticket: CS-4223

We've been sending the wrong value to the `CreatePrepaidCard` event - it should be the `issuer` instead of the `owner` (you can also see this in the [CreatePrepaidCard event definition](https://github.com/cardstack/card-pay-protocol/blob/main/contracts/PrepaidCardManager.sol#L53)).

This means the prepaid cards created via `PrepaidCardMarketV2` contract have the wrong `issuer` in the events, which means the subgraph has the wrong issuer stored in the database.

### DIFF
--- a/contracts/PrepaidCardManager.sol
+++ b/contracts/PrepaidCardManager.sol
@@ -52,6 +52,7 @@ contract PrepaidCardManager is Ownable, Versionable, Safe {
   event Setup();
   event CreatePrepaidCard(
     address issuer,
+    address owner,
     address card,
     address token,
     address createdFromDepot,
@@ -690,6 +691,7 @@ contract PrepaidCardManager is Ownable, Versionable, Safe {
     IERC677(token).safeTransfer(card, issuingTokenAmount - _gasFee);
 
     emit CreatePrepaidCard(
+      issuer,
       owner,
       card,
       token,
--- a/test/PrepaidCardMarketV2-test.js
+++ b/test/PrepaidCardMarketV2-test.js
@@ -788,6 +788,9 @@ contract("PrepaidCardMarketV2", (accounts) => {
         prepaidCardManager.address
       );
 
+      expect(createPrepaidCardEvent.issuer).to.be.equal(issuer);
+      expect(createPrepaidCardEvent.owner).to.be.equal(customer);
+
       let [provisionPrepaidCardEvent] = getParamsFromEvent(
         tx,
         eventABIs.PREPAID_CARD_MARKET_V2_PREPAID_CARD_PROVISIONED,
--- a/test/PrepaidCardMarket-test.js
+++ b/test/PrepaidCardMarket-test.js
@@ -724,12 +724,13 @@ contract("PrepaidCardMarket", (accounts) => {
   });
 
   describe("provision prepaid cards", () => {
-    let prepaidCards,
+    let createPrepaidCardTx,
+      prepaidCards,
       sku,
       askPrice = toTokenUnit(10);
 
     before(async () => {
-      ({ prepaidCards } = await createPrepaidCards(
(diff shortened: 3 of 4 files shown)


test/PrepaidCardMarketV2-test.js
```diff
@@ -788,6 +788,8 @@ contract("PrepaidCardMarketV2", (accounts) => {
         prepaidCardManager.address
       );
 
+      expect(createPrepaidCardEvent.issuer).to.be.equal(issuer);
```
jurgenwerk, line 791: This is the regression test - without the change in contracts/PrepaidCardManager.sol, this assertion failed, which is clearly wrong.

jurgenwerk:
After we deploy this, we need to adjust the subgraph mapping and rebuild it. Will take care of that.

contracts/PrepaidCardManager.sol
```diff
@@ -690,7 +690,7 @@ contract PrepaidCardManager is Ownable, Versionable, Safe {
     IERC677(token).safeTransfer(card, issuingTokenAmount - _gasFee);
 
     emit CreatePrepaidCard(
-      owner,
+      issuer,
```
FadhlanR, line 693: In the subgraph, I saw `handleCreatePrepaidCard` will mapping this issuer argument to prepaidCardEntity.issuer and prepaidCardEntity.owner, if we only update owner -> issuer in this argument then prepaidCardEntity.owner might have the wrong value. I think we should add issuer and owner in this event arguments.

jurgenwerk: Yes, I think it makes sense to add the owner to the event. Will add that change, thanks. I'm currently trying to wrap my head around how to change the mapping in the subgraph conditionally before & after this change so that we'll have historically accurate data

FadhlanR: Or we can create new event like CreatePrepaidCardV2

habdelra: it amounts a similar amount of work, but the downside here is that you increase the contract size and now have dead code. keep in mind contracts have a 24KB hard limit, and the PrepaidCardManager is probably already a bit on the large size. we should be very cautious around adding unnecessary bytes to the contract.

 the main challenge is that you need to use a different ABI in the subgraph for the older blocks vs the newer blocks. which is exactly what @jurgenwerk is planning on doing anyways. We've done this in the past, here is an example. The quirk is the cut over from the old to the new. The subgraph really doesn't like contracts that can be upgraded, and it's challenging to retain consistent data when you do this sort of thing. The main issue is that in your subgraph data source you need to tell the subgraph the block number to start indexing for the contract. For a contract that changes ABI, we need to know the block number that the contract was changed at before we actually make the change, basically a chicken and egg problem. to that end we have a short term solution to handle that scenario that is a tad inefficient (@jurgenwerk did a wonderful write up here: https://github.com/cardstack/cardstack/pull/3080). And the longer term scenario is to use the updated event shape along with a new datasource in our subgraph once we have the block number that the change was deployed in.

jurgenwerk: Thanks @FadhlanR for fixing the "stack too deep error", and @habdelra for thoroughly explaining the new datasource that will be needed in the subgraph! 

I squashed the commits into one and flagged this PR as "blocked" (by https://github.com/cardstack/cardstack/pull/3080). Please take another look if the code looks good to you and we can merge and deploy it once https://github.com/cardstack/cardstack/pull/3080 is shipped (and then use the block numbers to create a new datasource in the subgraph, and make the mapping rely solely on the event params).

FadhlanR (approved):
lgtm

PR #238, opened 2022-07-11
Create local blockchain nodes

FadhlanR:
Ticket: CS-4226

The idea is to create local blockchain nodes that already include the smart contract contracts needed to run protocol environments such as card payment protocol contracts, token bridges, and gnosis safe contracts.


alex-cardstack (approved):

jurgenwerk (approved):
I tested this locally and was able to connect to the node using hardhat and interact with the contracts. Nice job!

I think it would be good if you added a readme where you list commands you used to generated these zip files so that we can easily update them manually when we change the contracts, or even automatically later on (we could have a task that uploads the current snapshots to s3 and add a step to download it in the docker build step).

.openzeppelin/addresses-localhost.json
```diff
@@ -0,0 +1 @@
...":"RewardManager"},"RewardSafeDelegateImplementation":{"proxy":"0xB0D4afd8879eD9F52b28595d31B441D079B2Ca07","contractName":"RewardSafeDelegateImplementation"},"AddRewardRuleHandler":{"proxy":"0x922D6956C99E12DFeB3224DEA977D0939758A1Fe","contractName":"AddRewardRuleHandler"},"LockRewardProgramHandler":{"proxy":"0x1fA02b2d6A771842690194Cf62D91bdd92BfE28d","contractName":"LockRewardProgramHandler"},"RegisterRewardProgramHandler":{"proxy":"0x04C89607413713Ec9775E14b954286519d836FEf","contractName":"RegisterRewardProgramHandler"},"RegisterRewardeeHandler":{"proxy":"0x21dF544947ba3E8b3c32561399E88B52Dc8b2823","contractName":"RegisterRewardeeHandler"},"UpdateRewardProgramAdminHandler":{"proxy":"0xD8a5a9b31c3C0232E196d518E89Fd8bF83AcAd43","contractName":"UpdateRewardProgramAdminHandler"},"PayRewardTokensHandler":{"proxy":"0x51A1ceB83B83F1985a81C295d1fF28Afef186E02","contractName":"PayRewardTokensHandler"},"DAIUSDFeed":{"proxy":"0x8198f5d8F8CfFE8f9C413d98a0A55aEB8ab9FbB7","contractName":"ManualFeed"},"ETHUSDFeed":{"proxy":"0x0355B7B8cb128fA5692729Ab3AAa199C1753f726","contractName":"ManualFeed"},"CARDUSDFeed":{"proxy":"0x202CCe504e04bEd6fC0521238dDf04Bc9E8E15aB","contractName":"ManualFeed"}}
```
habdelra, line 1: @FadhlanR would it be possible to pretty print this JSON file like we do for the other address-*.json files? it makes it a littler easier to introspect that way

PR #239, opened 2022-07-14
Forbid registering a supplier twice

IanCal:
The test in commit https://github.com/cardstack/card-pay-protocol/commit/a2a2013c08cd3306c967d597ecbf4830832fdaa9 shows that if you register a supplier twice, a new depot is created and the contract loses track of the original one. I expect this would have bad side effects, while the funds in the original depot would not be lost the supplier manager would start returning a new empty safe.

I may be missing an expected flow though.

I've added an extra guard & test checking it's not able to be called twice.

### DIFF
--- a/contracts/SupplierManager.sol
+++ b/contracts/SupplierManager.sol
@@ -70,6 +70,7 @@ contract SupplierManager is Ownable, Versionable, Safe {
       msg.sender == bridgeUtils || (owner() == _msgSender()),
       "caller is not BridgeUtils nor owner"
     );
+    require(!isRegistered(supplier), "supplier already registered");
     return _registerSupplier(supplier);
   }
 
--- a/test/BridgeUtils-test.js
+++ b/test/BridgeUtils-test.js
@@ -121,6 +121,18 @@ contract("BridgeUtils", async (accounts) => {
     expect(await bridgeUtils.isRegistered(newSupplier)).to.equal(true);
   });
 
+  it("rejects a supplier registration if it has already been registered before", async () => {
+    let newSupplier = accounts[2];
+    // This is already registered from the previous test
+    expect(await bridgeUtils.isRegistered(newSupplier)).to.equal(true);
+
+    await bridgeUtils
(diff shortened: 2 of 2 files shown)


alex-cardstack (commented):
So, this call requires the sender is the bridge utils or the contract owner. We definitely don't do this from the owner on the regular, and the [bridge contract](https://github.com/cardstack/tokenbridge-contracts/blob/master/contracts/upgradeable_contracts/multi_amb_erc20_to_erc677/HomeMultiAMBErc20ToErc677.sol#L228-L233) checks if the supplier is registered and returns the existing safe if so.

I think this is a decent idea for defense in depth.

My only comment is that it's a bit unconventional to have a new `contract` block in the test instead of adding it into the existing contract block as a regular `it` statement, but that's nitpicking especially if it's a yak shave to make that work

IanCal:
> So, this call requires the sender is the bridge utils or the contract owner. We definitely don't do this from the owner on the regular, and the [bridge contract](https://github.com/cardstack/tokenbridge-contracts/blob/master/contracts/upgradeable_contracts/multi_amb_erc20_to_erc677/HomeMultiAMBErc20ToErc677.sol#L228-L233) checks if the supplier is registered and returns the existing safe if so.

Cool - do you think it's better to have a requires here or to have a similar behaviour (silently pass through)?

> My only comment is that it's a bit unconventional to have a new contract block in the test instead of adding it into the existing contract block as a regular it statement, but that's nitpicking especially if it's a yak shave to make that work

Ah yeah, I'd had it split out when I was testing it creating two as the tests all modify state here so when I added it it broke other tests. However now it's restricted with a requries, I can add the test in after the test that registration works.

I'm not overly fond of the tests interacting like that but I added in an expect at the beginning of the test and comment in case it breaks down the line by someone moving test order.

alex-cardstack (approved):
Looks good, I think here at the lower level, the require is fine and doesn't need to be idempotent, we'd just be doing the same thing twice

PR #241, opened 2022-07-22
Add README file to explain how to update zip files

FadhlanR:


jurgenwerk (changes requested):
Looking good. Can you also add the part where you generate the snapshots? Basically all the steps needed when we have to rebuild snapshots, including deploying all the contracts(list here: https://github.com/cardstack/cardstack/blob/main/packages/cardpay-sdk/contracts/addresses.ts#L41)

jurgenwerk:
I tested the new addresses file by deploying it locally to a local subgraph and it works great.

FadhlanR:
The instructions are already in README file

PR #248, opened 2022-09-02
Add a script that generates local addresses for the subgraph

jurgenwerk:
Ticket: CS-4337

This is a utility that will help with setting up the local protocol development environment. It generates locally deployed contract address in a format suitable for the phase where we generate the subgraph manifest file (subgraph.yml). The subgraph can then be deployed to the local graph node. More info in the code comment.


habdelra (approved):
looks good--thx!

